PR #1, opened 2021-09-21
Expand grammar for function signatures

hmontero1205:
Some notes about the changes I've made:

- Added detailed section about routine definitions and types to the LRM
- Updated the parser to match the routine definition grammar presented in the LRM
-  Hardcoded the compiler to only support single argument functions. That is, functions may either take a single variable or a single 1D tuple. This was meant to preserve compatibility with existing sslang code and also because further work must be done to the IR and the C runtime to support multi-dimensional tuples.
- Added frontend support for tuples, arrow types

- TODO: perform pass to unify function type annotations? -> vs :


doc/sslang-lrm.tex
```diff
@@ -328,6 +325,77 @@ \subsection{User-defined operators}
 a look at Agda's user-defined operators.
 
 \section{Types}
+sslang exposes two classes of types. \\
+
+\noindent The first class of types encompasses primitive types: integers, booleans, pure events (i.e. conditional variable), and unit (i.e. void).\\ 
+
+\noindent The second class of types are higher order types that are composed of primitive types:
```
j-hui, line 332: This description of type constructors seems to limit us only to the built-in type constructors, and does not permit user-defined type constructors.

Could we generalize and just say something indicating that type constructors are applied to zero or more type arguments, and that they must include the <above> built-in types?

also note that the "primitive" types are just type constructors that take no type arguments. I wouldn't consider them a different "class" of types, but it is still worth giving examples of them to clarify that these types do exist and are quite common/important.

doc/sslang-lrm.tex
```diff
@@ -328,6 +325,77 @@ \subsection{User-defined operators}
...
+\noindent This is summarized by the following grammar:
+\setlength{\grammarindent}{5em}
+\begin{grammar}
+<type> ::= [<type>]
+\alt (<type>(, <type>)+)
+\alt <type> $\rightarrow$ <type>
+\alt \&<type>
+\alt Sched <type>
+\alt (<type>)
+\alt Int8 | Int16 | Int32 | Int64
```
j-hui, line 350: These type names should all be encompassed under a single rule for typenames (and the same should go for `Sched`).

So something like
```
<type> ::= [<type>]
    | (<type>(, <type>)+)
    | <type> -> <type>
    | &<type>
    | (<type>)
    | ()
    | <type-id>

<type-id> ::= [A-Z][a-zA-Z0-9]*
```

doc/sslang-lrm.tex
```diff
@@ -328,6 +325,77 @@ \subsection{User-defined operators}
...
+    \item Postfix-tuple: Type annotations for tuples are written after the tuple, as in OCaml.\\\\
+    \texttt{foo ((a, b): (Int, Bool)) $\rightarrow$ Int = \{...\}}
+    \item Postfix-functional: Haskell-like type signature expressed after all parameters. For example, a function that takes two integers and returns a boolean:\\\\
+    \texttt{foo x y: Int $\rightarrow$ Int $\rightarrow$ Bool = \{...\}}
+\end{itemize}
+
+\noindent These forms are summarized by the following grammar:
+\setlength{\grammarindent}{5em}
+\begin{grammar}
+<FUNCDEF> ::= ID () ($\rightarrow$ <TYPE>)? = \{ <BODY> \}
```
j-hui, line 387: Is there a reason the terms in the BNF grammar are now capitalized?

doc/sslang-lrm.tex
```diff
@@ -328,6 +325,77 @@ \subsection{User-defined operators}
...
+\begin{itemize}
+    \item Routine name
+    \item Zero or more parameters, each of which are optionally typed
+    \item Optionally specified return type
+    \item Routine body
+\end{itemize}
+\noindent The compiler will perform type inference for untyped components and will verify that typed components are correct.
+
+\subsection{Routine Signatures}
+\noindent sslang supports three ways of writing out type signatures for routines.
```
j-hui, line 368: I think stating that there are just three ways doesn't account for the variations of functions which take further-nested tuples, i.e., more tuple arguments. For instance, `f ((a: T, b: T), c: T) -> T`, `f (((a, b): (T, T)), c: T) -> T`, `f (((a, b), c): ((T, T), T)) -> T`, and `f ((a, b), c): ((T, T), T)` are all valid way of declaring the equivalent type signature for `f`, and should all be valid in the grammar (though in practice only the first and last are readable).

My suggestion is to base the description off of the following points:
- Our type annotation syntax uses `:` to mean "is of type", and `->` to mean "returns".
- Each function declaration must specify the types of all its arguments, and its return type.
- There are two broad syntactic categories of type signatures: those that look like `f <args> : <fn-type>`, and those that look like `f <typed-args> -> <ret-type>`. The former syntax is evocative of functional languages like OCaml and Haskell, while the latter is evocative of languages like Python and Rust.
- When type-annotating `<typed-args>`, we use `:` to annotate arguments. Our syntax supports "distributing" the `:` over tuple members, so writing `((a, b): (A, B))` is equivalent to `(a: A, b: B)`.
- No identifier should be type-annotated more than once.

This should cover the idea of how to write a syntactically valid type signature for a routine. Those who are interested in working out all the permutations of the above rules may do so while inspecting the provided grammar.

j-hui (commented):
It looks like you're still working on this, so I'll just comment on what's already here. This is a great start, but I think there are some details to be brushed up.

Also, I think we should clean out our notes from the wishlist embedded in the LRM, especially as the LRM sections they discuss are already coming to fruition.

One last trailing thought is, should we leave generic the syntax for type operators? For instance, to allow a user to define their own symbolic type operators, e.g., `?T` or `A + B` as types? There are limited cases where this may be useful/illustrative, yet countless cases where this is aesthetically alarming.

j-hui:
Another thought: this grammar for type expressions does not yet account for (1) type variables, and (2) type predicates (i.e., type classes). Type variables are simple enough to include, following the lead of what Haskell does, though I'm not the biggest fan of the syntax other languages have for type predicates/constraints

doc/sslang-lrm.tex
hmontero1205: Yup! I realized I'd have to rewrite this section after the Slack convo today.

hmontero1205:
> Another thought: this grammar for type expressions does not yet account for (1) type variables, and (2) type predicates (i.e., type classes). Type variables are simple enough to include, following the lead of what Haskell does, though I'm not the biggest fan of the syntax other languages have for type predicates/constraints

I think that the goal of this PR was to expand the grammar for function definitions. As a result, I had to work with type expressions a bit, but they were never my focus. I think further work on type expressions should be in a separate PR.

doc/sslang-lrm.tex
hmontero1205: > Each function declaration must specify the types of all its arguments, and its return type.

I'm not exactly sure about this, especially with type inference in the pipeline. It was to my understanding that whatever isn't annotated will be filled in and whatever is annotated will be verified by the type checker. As such, the grammar makes type annotations optional.

j-hui: We haven't worked out the details or extent of the type system just yet. Depending on what flavor of type inference we do, we may be forced to annotate the type (for some type systems, esp those involving type classes, inference is undecidable). We can easily relax this requirement once that's worked out. 

Also, even if we do have inference capabilities, I still think it's a good idea to annotate types for top-level functions. Even if they're not needed by the type checker, they perform an important role of conveying the interface of a function to programmers.

doc/sslang-lrm.tex
```diff
@@ -328,6 +233,84 @@ \subsection{User-defined operators}
...
+    \item Zero or more parameters, each of which are optionally typed
+    \item Optionally specified return type
+    \item Routine body
+\end{itemize}
+\noindent The compiler will perform type inference for untyped components and will verify that typed components are correct.
+
+\subsection{Routine Signatures}
+\noindent At the core of sslang routine signatures, we have two symbols that provide type information about a language construct. Their meanings are as follows:
+\begin{itemize}
+    \item \texttt{:} (Colon): ``is of type"
```
j-hui, line 279: Can you give an example of the two different forms upfront? 

In fact, I think the "two broad categories" could be stated earlier, since it outlines the top-down structure for the reader to understand the rest of the section.

j-hui (commented):
I haven't yet closely looked at the Ast.hs and Ast2Ir.hs changes yet, but the work on the LRM mostly lgtm. There are still some things that will need to be fleshed out once we work out more details.

I'll refrain from approving just yet and stamp it later once #3 is approved and merged in (or perhaps this should be switched to merge into that branch instead?).

doc/sslang-lrm.tex
hmontero1205: The issue is the two forms of tuple annotation. I made the type annotations optional because it simplifies the grammar so much when dealing with tuples. The idea there was that a subsequent pass would verify existence of annotations and only one annotation per variable. I can change it to be mandatory though. Please let me know.

hmontero1205:
fyi @j-hui I decided to use `main` as my base, not your wip type inference branch. I don't use any of your code and I think what I have here is a nicely contained change.

hmontero1205:
OK @j-hui , I've just added Edwards' suggestions from our last meeting. This PR got really big so if you want to call to go over this, LMK.

j-hui:
Thanks Han! Why don't we go over it today in our meeting?

src/Ast.hs
```diff
@@ -22,12 +22,17 @@ type OperatorId = String
...
 
-data Bind = Bind [VarId] Ty
+data Declaration = Function VarId [Bind] Expr FnTyAnnotation
+
+data Bind = Bind VarId (Maybe Ty)
+          | TupBind [Bind] (Maybe Ty)
 
 data Ty = TCon TConId
         | TApp Ty Ty
+        | TTuple [Ty]
```
hmontero1205, line 35: Hardcode TArrow

src/Ast.hs
```diff
@@ -22,12 +22,17 @@ type OperatorId = String
...
 data Program = Program [Declaration]
 
-data Declaration = Function VarId [Bind] Expr
+data FnTyAnnotation = ReturnType Ty
+                    | CurriedType Ty
 
-data Bind = Bind [VarId] Ty
+data Declaration = Function VarId [Bind] Expr FnTyAnnotation
+
+data Bind = Bind VarId (Maybe Ty)
```
hmontero1205, line 30: Make VarId be optional, allow something like `foo () () ()`, `foo _ _ _ `

hmontero1205: @j-hui , making it optional means losing whether it was a () or a _. I am just making the `VarId` whatever is matched for now.

hmontero1205:
@j-hui rebsaed onto main, ready for review after your round of feedback on friday.

src/Ast.hs
j-hui: Yeah this makes sense. I won't make this a blocker for this PR, but I think `Bind` should eventually resemble something like what is found in the pattern of a pattern-match arm.

j-hui (approved):
lgtm

PR #2, opened 2021-09-21
Ignore cabal file to avoid unnecessary conflicts

j-hui:
Tracking the .cabal file in version control is supposed to be a best practice, but in practice it seems to just cause a lot of merge conflicts/version-control noise, and can lead to unexpected builds as a result of stale cabal files.

So, in spite of best practice, we will ignore this for now, and then include it once we get to the point where we can "release" this compiler.


j-hui:
Going to violate my own branch protection rules here and merge without review, since I already cleared this with @sedwards-lab .

PR #7, opened 2021-10-04
Define IR, define type systems, and implement Codegen

j-hui:
- Define a typesystem-parameterized expression language IR for SSM
- Define built-in types needed by codegen
- Define a flat and polymorphic type system which wrap those built-in types
- Compile IR to C using language-c-quote


j-hui:
Note that the codegen phase leaves a few major language features/compiler passes unimplemented:
- user-defined data types (@EmilySillars)
- scope analysis and memory management (@Scanteianu)
- type inference and monomorphization (@XijiaoLi + @LeoQiao18)
- lambda lifting + defunctionalization

The design and implementation of type classes is also as of yet unattempted.

Right now, the AST -> IR lowering phase is not yet implemented, but I will work on that once I merge this work with @hmontero1205 and @maxhhelman 's most recent changes to the language frontend.

sedwards-lab (approved):
LGTM

PR #8, opened 2021-10-06
Validate routine type signatures

hmontero1205:
That is,
- Assert all args are annotated exactly once for Pythonic signatures
- Assert all args are not annotated and that the annotated arrow type is consistent with the number of args for Haskell-like signatures

### DIFF
--- a/src/Passes.hs
+++ b/src/Passes.hs
@@ -2,16 +2,13 @@ module Passes where
 
 import qualified Ast as A
 
-checkRoutineSignatures :: A.Program -> A.Program
-checkRoutineSignatures (A.Program decls) = A.Program $ map checkAnnotations decls
+checkRoutineSignatures :: A.Program -> Bool 
+checkRoutineSignatures (A.Program decls) = and $ map checkAnnotations decls
   where
-    checkAnnotations decl@(A.Function _ binds _ fnTyAnn) =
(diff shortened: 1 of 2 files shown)


src/Passes.hs
```diff
@@ -0,0 +1,24 @@
+module Passes where
+
+import qualified Ast as A
+
+checkRoutineSignatures :: A.Program -> Bool 
+checkRoutineSignatures (A.Program decls) = all checkAnnotations decls
```
j-hui, line 6: Please write a haddock comment to document what this is checking.

j-hui (changes requested):
Lgtm, but there are some organizational changes I'd like to suggest:

- rename `Passes.hs` to `Lint.hs` or `Checks.hs` or something. Everything operational component of a compiler can be thought of as a compiler pass so that isn't very descriptive imo. 
- move this checking code, as well as the scanner, parser, and AST definition into a new subdirectory `src/Frontnew/`. I would like to clarify the boundary between the different phases of the compiler. 
- other nits noted inline. 

At some point in the future, we'll also really want to scale out the checking while keeping error messages reporting consistent. So for now we can return Bool but that should really return MonadExcept CompilerError () or something like that.

j-hui:
In that case, don't worry about Main.hs for now. We'll do another PR where
we just clean up the entire repo and make formatting consistent everywhere.

PR #10, opened 2021-10-07
Pretty-print IR

Scanteianu:
Implement basic IR pretty printing

### DIFF
--- a/src/Common/Pretty.hs
+++ b/src/Common/Pretty.hs
@@ -12,7 +12,7 @@ module Common.Pretty
   ) where
 
 import           Prettyprinter
-import           Prettyprinter.Internal
+import           Prettyprinter.Render.String
 
 -- | @typeAnn t d@ annotates document @d@ with type annotation @t@.
 typeAnn :: Pretty t => t -> Doc ann -> Doc ann
@@ -44,5 +44,5 @@ block separator = braces . sep . punctuate separator
 
 -- | Format with a document of infinite width, preventing wraparound.
 spaghetti :: Pretty t => t -> String
-spaghetti p = renderShowS (layoutPretty opts (pretty p)) ""
-  where opts = LayoutOptions { layoutPageWidth = AvailablePerLine 200 1.0 }
+spaghetti = renderString . layoutPretty opts . pretty
+  where opts = LayoutOptions { layoutPageWidth = Unbounded }


sedwards-lab:
Something you might consider, which would be even more useful: make the IR pretty printer spit out syntactically correct Sslang code. It might be a bit ugly, but it would be interesting to be able to send the result back through the compiler. We could use this during testing, for example.

README.md
```diff
@@ -1,5 +1,9 @@
 # sslang
 
+**NOTE: Development for this project has moved to [ssm-lang/sslang](github.com/ssm-lang/sslang).**
```
j-hui, line 3: Please rebase this commit out of the PR

src/IR/IR.hs
```diff
@@ -252,3 +255,97 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
   wfPrim (PrimOp PrimLt    ) [_, _]    = True
   wfPrim (PrimOp PrimLe    ) [_, _]    = True
   wfPrim _                   _         = False
+
+
+{-
+how to use:
```
j-hui, line 261: These command line usage notes should be removed from documentation inline, esp the questions further below.

src/IR/IR.hs
```diff
@@ -252,3 +255,97 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+show (LitIntegral 3)
+show New
+Questions:
+why does `show (Lit "a" Integer)` not work? -- b/c types not defined, just use () as a type
+why does `show (Var (VarId (Identifier "foo")) LitIntegral)` not work?
+
+show (Var (VarId (Identifier "foo")) ()))
+
+-}
+instance (Pretty t) => Show (Expr t) where
```
j-hui, line 274: The `Show` instance should be derived, and the `Pretty` instance should be kept separate.

In my mind, `Show` should really dump the Haskell representation of a data structure (and it should be the inverse of `Read`).

Meanwhile, `pretty` should be the inverse of `parse`, modulo compiler transformations

src/IR/IR.hs
```diff
@@ -252,3 +255,97 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+    pretty (LitIntegral i) = pretty (show i)
+    pretty (LitBool b) = pretty (show b)
+
+instance Pretty Primitive where
+  pretty New = pretty "new"
+  pretty Dup = pretty "dup"
+  pretty Drop = pretty "drop"
+  pretty Reuse = pretty "reuse"
+  pretty Deref = pretty "deref"
+  pretty Assign = pretty "="
```
j-hui, line 296: Assign should denoted by an infix `<-`; something like `Assign x y` should be `x <- y`.

You may want to consider not having `Primitive` as an instance of `Show`, and instead matching on each enum explicitly in `pretty (Prim ..)` in `instance Pretty Expr t where ...`.

src/IR/IR.hs
```diff
@@ -252,3 +255,97 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+  pretty PrimGe = pretty ">="
+  pretty PrimLt = pretty "<"
+  pretty PrimLe = pretty "<="
+
+instance (Pretty t) => Pretty (Expr t) where
+  {- do pretty show(VarId)?
+  this might be needed if other file uses legacy printer
+  var myInt LitIntegral?
+
+  //do things like varId take args?
```
j-hui, line 327: I don't think I fully understand this question

src/IR/IR.hs
```diff
@@ -252,3 +255,97 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+
+  //do i have to wrap each expression in parens somehow?
```
j-hui, line 329: this is... extremely lispy :eyes: 

You can see the `Pretty` instance for the `Ast` as it is right now for how some of these constructs should look; the AST consists of, roughly speaking, a superset of language constructs found here in the IR.

unit-tests/IRTest.hs
```diff
@@ -0,0 +1,28 @@
...
+import Data.List
+
+{-
+How to run:
+in src, run `stack ghci`
+then do ``:l unit-tests/IRTest.hs`
+then do ``:main`
+
+-}
+tests  = [(show(New),"new"), (show (Var (VarId (Identifier "foo")) ()),"(var foo () )")]
```
j-hui, line 14: This is fine for now, but we should really set up hspec for unit and regression tests

src/IR/IR.hs
```diff
@@ -252,3 +259,117 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+  // this feels v. lispy
+  -}
+  pretty (Var a b) =
+    pretty "(var" <+> pretty (show a) <+> pretty b <+> pretty ")"
+  pretty (Data a b) =
+    pretty "(data" <+> pretty (show a) <+> pretty b <+> pretty ")"
+  pretty (Lit a b) = pretty "(literal" <+> pretty a <+> pretty b <+> pretty ")"
+  pretty (App a b c) =
+    pretty "(apply" <+> pretty a <+> pretty b <+> pretty c <+> pretty ")"
```
hmontero1205, line 342: this should parse into something like `(a) (b) : c`

src/IR/IR.hs
```diff
@@ -252,3 +259,117 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+  //do things like varId take args?
+
+  //do i have to wrap each expression in parens somehow?
+  // this feels v. lispy
+  -}
+  pretty (Var a b) =
+    pretty "(var" <+> pretty (show a) <+> pretty b <+> pretty ")"
+  pretty (Data a b) =
+    pretty "(data" <+> pretty (show a) <+> pretty b <+> pretty ")"
+  pretty (Lit a b) = pretty "(literal" <+> pretty a <+> pretty b <+> pretty ")"
```
hmontero1205, line 340: `(a: b)`

src/IR/IR.hs
```diff
@@ -252,3 +259,117 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+  pretty (App a b c) =
+    pretty "(apply" <+> pretty a <+> pretty b <+> pretty c <+> pretty ")"
+  pretty (Let a b c) =
+    pretty "(let" <+> pretty (show a) <+> pretty b <+> pretty c <+> pretty ")"
+  pretty (Lambda a b c) =
+    pretty "(lambda" <+> pretty (show a) <+> pretty b <+> pretty c <+> pretty
```
hmontero1205, line 346: `((\ a -> b): c)`

src/IR/IR.hs
j-hui: No and yes and no. 

No because the scanner is currently inserting braces in a way that breaks examples with matching parens. 

Yes because assuming the scanner is working, inserting parens defensively is the simplest way to get around precedence issues when "pretty"-printing. 

No because what you end up printing isn't actually very pretty, and there should be ways to use the pretty-printer monoid to only insert parens where needed.

Scanteianu: starting with parens seems safer - will do that for now

hmontero1205: I agree, fixing the frontend scanner seems like a me issue :)

src/IR/IR.hs
```diff
@@ -252,3 +259,117 @@ wellFormed (Prim p es _) = wfPrim p es && all wellFormed es
...
+      ")"
+  pretty (Match a b c d) =
```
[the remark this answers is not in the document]
Scanteianu: in 

`Match (Expr t) Binder [(Alt, Expr t)] t` 

what is the binder - should I show it? is it the x above?  

is the t ": int" above?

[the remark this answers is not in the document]
Scanteianu: i think I have it

sedwards-lab:
Did this facility ever get linked to the cli? E.g., does the --dump-ir directive use this code?

j-hui:
Yeah, here: https://github.com/ssm-lang/sslang/pull/10/files#diff-d3e498a88b4c374b1edbd8693542a73e944d51a24a64a683658926ba655bc6ffR122 .

`spaghetti` is a ~pretty~ printer that just dumps everything on one line, and not meant to stick around; I hope to get rid of it once we print things with whitespace.

One thing I haven't really thought about yet is, _which_ IR should this dump directive spit out? There's at least 4 involved, and more if we consider all the intermediate passes.

PR #20, opened 2021-10-15
Expand scanner test suite

hmontero1205:
Things to figure out:
- Inline loops
- Inconsistent use of `TThen`
- Explicit braces around bind group doesn't lex
- Extra braces in explicit braces around definition

### DIFF
--- a/test/scanner/Tests/ScanConditionalsSpec.hs
+++ b/test/scanner/Tests/ScanConditionalsSpec.hs
@@ -14,13 +14,53 @@ import           Front.Token                    ( TokenType(..) )
 spec :: Spec
 spec = do
   it "scans inline conditionals" $ do
-    pendingWith "help needed"
+    let input = "if a { b }"
+        output =
+          [ TIf
+          , TId "a" -- TODO(hans): What happened to TThen?
+          , TLbrace
+          , TId "b"
+          , TRbrace
+          ]
+    scanTokenTypes input `shouldBe` Right output
 
   it "scans conditionals blocks" $ do
-    pendingWith "help needed"
+    let input = unlines
+          [ "if a"
+          , "  b"
+          ]
+        output =
+          [ TIf
+          , TId "a"
+          , TThen
+          , TLbrace
+          , TId "b"
+          , TRbrace
+          ]
+    scanTokenTypes input `shouldBe` Right output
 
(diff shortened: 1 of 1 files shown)


test/scanner/Tests/ScanLetSpec.hs
```diff
@@ -95,32 +137,121 @@ spec = do
...
+          ]
+    scanTokenTypes input `shouldBe` Right output
 
   it "supports explicit braces around bind group" $ do
     pendingWith "help needed"
     -- let { a = b
     --   and b = a }
     --
     -- though this looks rather unsightly
+    -- TODO(hans): The scanner isn't actually OK with this
```
j-hui, line 173: I don't think the scanner is looking for the `and` token, that might be the problem.

test/scanner/Tests/ScanLetSpec.hs
```diff
@@ -95,32 +137,121 @@ spec = do
...
+          , "            a_block_2_2 }"
+          , "outside_let"
+          ]
+        output =
+          [ TLet
+          , TLbrace
+          , TId "a"
+          , TEq
+          , TLbrace
+          , TLbrace -- Inserts extra brace pair?
```
j-hui, line 153: The extra brace pair is due to the fact that, at the moment, `=` starts a block + there's a bug in the scanner where it doesn't check if the user already explicitly started a block before starting one itself.

Technically not an issue on its own because `{x}` is just parsed as `x` now, but there are further problems where the scanner doesn't actually end the blocks it starts at the right place.

test/scanner/Tests/ScanConditionalsSpec.hs
```diff
@@ -14,13 +14,53 @@ import           Front.Token                    ( TokenType(..) )
 spec :: Spec
 spec = do
   it "scans inline conditionals" $ do
-    pendingWith "help needed"
+    let input = "if a { b }"
+        output =
+          [ TIf
+          , TId "a" -- TODO(hans): What happened to TThen?
```
j-hui, line 20: Great question. I think when it saw `{`, it (correctly?) decided not to insert the additional block.

j-hui (approved):
lgtm, feel free to just merge (esp since we're not merging into main; my branch is itself a wip anyway).

PR #24, opened 2021-10-19
Fix pretty-printer for refactored AST

hmontero1205:


src/Front/Ast.hs
```diff
@@ -108,88 +110,69 @@ data Literal
...
+    pretty pat
+      <+> pretty ": ["
+      <+> hsep (punctuate comma $ map pretty tys)
+      <+> pretty "]"
+
+instance Pretty BindPat where
+  pretty BindWildcard   = pretty '_'
+  pretty (BindId  s   ) = pretty s
+  pretty (BindLit l   ) = pretty l
+  pretty (BindAs b p  ) = pretty b <> pretty '@' <> pretty p -- TODO(hans): Parens around p?
```
j-hui, line 147: Yeah you may as well defensively add parens around everything.

We will eventually take another pass that adds precedence to things.

src/Front/Ast.hs
```diff
@@ -108,88 +110,69 @@ data Literal
...
+  pretty (Lit l              ) = pretty l
   pretty (Apply    (Id id) e ) = pretty id <+> pretty e
   pretty (Apply    e1      e2) = parens (pretty e1) <+> pretty e2
   pretty (OpRegion e1      r ) = parens (pretty e1 <> p r)
    where
     p EOR             = emptyDoc
     p (NextOp s e r') = space <> pretty s <+> pretty e <> p r'
-  pretty (As v e) = pretty v <> pretty '@' <> pretty e
-  pretty NoExpr   = emptyDoc
+  pretty NoExpr = emptyDoc
```
j-hui, line 175: This should either throw an error, or chuck out some very visible `pretty "NOEXPR"` token here. `pretty` shouldn't be called on `NoExpr` on its own, because it really shouldn't appear anywhere aside from in the `else` branch of the `IfElse` construct.

j-hui: My concern with just returning an empty document here is that, from a malformed AST, you will produce a malformed document that doesn't parse because of a quietly missing token, and it will be annoying to debug. So if you come across such a malformed AST, it should either fail early or fail loudly.

src/Front/Ast.hs
```diff
@@ -209,23 +192,20 @@ instance Pretty Expr where
       <+> pretty v
       <+> pretty "<-"
       <+> pretty e2
-  pretty (Assign v e) = pretty v <+> pretty "<-" <+> pretty e
+  pretty (Assign     v e) = pretty v <+> pretty "<-" <+> pretty e
```
j-hui, line 195: There should be braces wrapping what comes after `<-` and `=`.

test/prettyprinter/Tests/PrettyAstSpec.hs
```diff
@@ -0,0 +1,48 @@
...
+                                                )
+
+import           Front                          ( renderAst )
+import           Front.Ast
+import           Front.Parser                   ( parseProgram )
+
+parseAndRender :: String -> String
+parseAndRender input = case parseProgram input of
+  Left  e   -> error e
+  Right ast -> renderAst ast
```
hmontero1205, line 16: @j-hui do you think this is a reasonable way to do this? Am I missing some high IQ trick to lift the operations into the `Either` monad or something instead of doing this unwrapping?

j-hui: If you are so inclined, you could write this point-free as:

```haskell
parseAndRender :: String -> String
parseAndRender = bimap error renderAst . parseProgram
```

where `bimap` is a method of the `Bifunctor` type class (you'll need to import it).

j-hui:
With @hmontero1205 permission, I modified the tests to actually compare ASTs rather than strings. This revealed some precedence issues with the pretty-printer, which I fixed by defensively parenthesizing everything (except the lower precedence constructs).

PR #25, opened 2021-10-19
Syntax fixes

j-hui:
Draft PR for now, until the following are done:
- [ ] Merge in #24 
- [ ] Get syntax and scanner details into LRM

But most parts should be ready to look at now. Changes wrt `main`, at a high level, are:

- Drop `then`, `and`, `new`, and `of` keywords.
- Change `case` to `match` in the lexer (though there is yet to be a `Match` statement in the AST, nor any `match` expression in the grammar, pending @maxhhelman 's work)
- Expand responsibility of scanner to do delimiter matching, which is important for knowing how many implicit braces to insert in the presence of explicit delimiters, thus fixing prior issues with parenthesization working unexpectedly
- Introduce hspec for unit-testing scanner and parser, and begin to populate test suites (will continue to build upon this as I continue to work on this)
- Consolidate on a uniform representation of "patterns" in both the syntax and the AST, i.e., "that which appears on the LHS of let-bindings, lambdas, and pattern match arms".
- General rewrite of the scanner and parser, to add in lots of documentation and to clarify what some things are doing


sedwards-lab:
Make sure this operation can be done through a standalone executable. We'll run this test on every program in the test suite. At the very least, add "pretty-print the AST" as an option for sslc so that we can parse and pretty print each example, then parse and pretty print the result and verify that the two pretty-printed outputs are identical. The reason to do it this way is to make it easier to diagnose the problem (diff output is a lot easier to understand).

j-hui:
@sedwards-lab do you have any suggestions for how I present the scanner's token insertion algorithm? I've just written it out as a plaintext decision tree though I think it could really be improved. Also, any general feedback about the LRM writing is greatly welcome; I erred on the side of writing more than less so let me know if there are bits that are too wordy.

PR #34, opened 2021-10-23
Add syntax for pattern matching

maxhhelman:
Still need to generate IR from this


src/Front/Parser.y
```diff
@@ -35,10 +35,12 @@ import Common.Compiler (ErrorMsg)
   'let'   { Token (_, TLet) }
   'after' { Token (_, TAfter) }
   'wait'  { Token (_, TWait) }
-  'fun'   { Token (_, TFun) }
```
j-hui, line 38: Have you merged from `main` recently? This should not have been deleted. Please do not merge until this is fixed.

src/IR/IR.hs
```diff
@@ -232,12 +232,13 @@ instance Comonad Expr where
   extend f e@(Prim p es _) = Prim p (fmap (extend f) es) (f e)
 
 instance Foldable Expr where
-  foldMap f (Var  _ t     ) = f t
-  foldMap f (Data _ t     ) = f t
-  foldMap f (Lit  _ t     ) = f t
-  foldMap f (App    l  r t) = foldMap f l <> foldMap f r <> f t
-  foldMap f (Let    xs b t) = mconcat (map (foldMap f . snd) xs) <> foldMap f b <> f t
-  foldMap f (Lambda _  b t) = foldMap f b <> f t
+  foldMap f (Var  _ t ) = f t
```
j-hui, line 235: As far as I can tell, this is just a formatting change (might have been introduced by mistake); please revert this

src/IR/LowerAst.hs
```diff
@@ -133,6 +137,16 @@ lowerExpr (A.IfElse c t e) k = I.Match cond Nothing [tArm, eArm] (k I.untyped)
...
 
+-- | Lower an A.Pat into an I.Alt
+lowerAlt :: A.Pat -> I.Alt
+lowerAlt A.PatWildcard  = I.AltDefault
+lowerAlt (A.PatId  _  ) = error "I.Alt for A.PatID not implemented yet"
+lowerAlt (A.PatLit l  ) = I.AltLit $ lowerLit l
+lowerAlt (A.PatTup _  ) = error "I.Alt for A.PatTup not implemented yet"
+lowerAlt (A.PatCon _ _) = error "No way to lower A.DataCons to I.DataCons yet"
+lowerAlt (A.PatAnn _ p) = lowerAlt p
+lowerAlt (A.PatAs  _ p) = lowerAlt p
```
j-hui, line 148: You should throw an error here; something of the form `match x { a@b => { e }` should be desugared to `match x { b => { let a = b ; e }`. But that should all be part of more extensive desugaring, and shouldn't be handled here for now.

j-hui (changes requested):
There are still a few things to fix---a couple of them are just things to clean up, but I'm marking this as `Request changes` because I want to make sure we don't accidentally remove lambdas from the language when we merge to main...

In the meantime, @maxhhelman I'll make a quick PR implementing some of the changes to the IR, to make your job with `lowerAlt` a little easier.

src/IR/IR.hs
j-hui: actually nvm; if brittany decided to change this, then let's leave it like this (since brittany is supposed to be idemopotent). It'll be clobbered sooner or later.

src/Front/Parser.y
j-hui: nvm I just fixed this

src/IR/LowerAst.hs
```diff
@@ -123,16 +123,30 @@ lowerExpr (A.Wait exprs) k =
...
   tArm = (I.AltLit (I.LitBool True), lowerExpr t id)
-  eArm = (I.AltDefault, lowerExpr e id)
+  eArm = (I.AltDefault Nothing, lowerExpr e id)
 lowerExpr (A.OpRegion _ _) _ = error "Should already be desugared"
 lowerExpr A.NoExpr         k = I.Lit I.LitEvent (k I.untyped)
 
+-- | Lower an A.Pat into an I.Alt
+lowerAlt :: A.Pat -> I.Alt
+lowerAlt A.PatWildcard  = I.AltDefault Nothing
+lowerAlt (A.PatId  _  ) = error "I.Alt for A.PatID not implemented yet"
```
j-hui, line 143: This can now be directly translated to the IR, using `IR.AltDefault (Just v)` wher `v` is the id in `PatId`.

j-hui (approved):
I was _going_ to make a PR into this on another branch when fixing the IR `Match` statement... but I inadvertently pushed to this branch. So I decide to address some of the other logistical issues while I was at it.

There are still a couple of nits with the `lowerAlt` function, but it's not being used yet by any regression test so I don't consider those blocking issues. (If there's any problems, I'll fix them when implementing pattern matching in codegen.) Feel free to merge once you're done with those (and make sure tests continue to pass).

PR #39, opened 2021-10-27
Add sections for expressions in LRM

hmontero1205:
Outstanding items:

- [x] Format and Style (I chose grammar -> blurb -> example)
- [x] Lambdas
- [x] Pattern-matching (leaving as TODO)
- [x] Op Regions (leaving as TODO)


j-hui (approved):
What's written LGTM so far. Some stuff could be refined and there are still parts missing (namely, pattern matches and op regions), but let's get this in the `main` branch for now.

PR #40, opened 2021-10-27
Update type inference to support basic typing context

XijiaoLi:
In this PR:
- Implemented typing context `Context` to keep a record of all the type information (for now, it's just all the type annotations we have seen). In this way, we now support minimal type annotations.
- Added `typeinference` test suite with some basic cases.
  - In order to use `HSpec.shouldBe`, I added a bunch of `deriving Eq` to support comparison between `Expr`.

All unit tests are passing.


XijiaoLi:
From this point forward, we will switch to using Monad for implementing type inference. Currently I still haven't figure out the cause of the mysterious `Prelude.head: empty list` error. Inside the unit test, I dumped the inference result of `doubleblink.ssl` and it looks fine (all the expressions have >= 1 annotated types). To fix this and at least compile `doubleblink.ssl` without error, I added an dummy `Error` type under `data Builtin t` to somehow indicate this empty list error. I will leave it as is for now; we can discuss it more during tomorrow's meeting.

j-hui (commented):
Awesome start, this is already looking much better than what hideous implementation I had before. I'm neither going to approve nor request changes since I'm not sure if this is still a WIP (are tests passing?)

I left a bunch of suggestsions throughout; we can go through them in more detail later during our meeting.

Another organization suggestion is to lift out the inference rules for `I.Prim` and `I.Lit`, and keep those at the top-level (i.e., `inferPrim` and `inferLit`), rather than folding them in with `inferExpr` and doing nested pattern matching like I had done before. This is purely so that it's easier to navigate, but ideally we should be able to easily distinguish "here are the typing rules for the core lambda calculus" and "here are the typing rules for our primitive extensions for the lambda calculus"

regression-tests/tests/doubleblink2.ssl
```diff
@@ -0,0 +1,20 @@
+toggle(led : &Int) -> () =
+  led <- (1 - deref led)
```
j-hui, line 2: You should be able to remove the parens around `1 - deref led`

src/Common/Identifiers.hs
```diff
@@ -119,6 +119,7 @@ newtype FfiId = FfiId Identifier
 -- | ToIdentifier for user-defined variable, e.g., @x@
 newtype VarId = VarId Identifier
   deriving Eq
+  deriving Ord via Identifier
```
j-hui, line 122: While you're at it, can you add this to all other identifier types? They are likely candidates for being keys of maps.

j-hui (changes requested):
Looks good to me overall! Three more small changes before merging:

- make `inferPrim` its own function and extract that from `inferExpr`. It should be functionally the same but will just make `inferExpr` a little easier to navigate. (To be clear, `inferPrim` doesn't exist yet and is currently interleaved in `inferExpr`.)
- get rid of the match-all pattern in `inferPrim` and throw an error if you come across a primitive you don't know how to type. 
-  make `anns2Class` monadic, and have it throw an internal compiler error if the type list is empty, and a type error if there are any issues unifying the type list.

src/IR/TypeInference.hs
```diff
@@ -19,57 +22,144 @@ import           IR.Types.TypeSystem            ( dearrow
...
+localInferExpr [] = return []
+localInferExpr ((b, e):vs) = do
+  e' <- inferExpr e
+  case b of
+    Just vid -> insertVar vid (extract e')
+    Nothing -> return ()
+  vs' <- localInferExpr vs
+  return $ (b, e'):vs'
+
+anns2Class :: Ann.Type -> Classes.Type
```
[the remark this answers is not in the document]
XijiaoLi: Turns out this is the hardest part.. lol But I've figured it out!

src/IR/TypeInference.hs
```diff
@@ -4,72 +4,203 @@ For now, this module implements a very crude form of pure, syntax-directed type
...
+      vty' <- anns2Class vty
+      insertVar v vty'
+    Nothing -> throwError $ Compiler.TypeError $ "Lambda wasn't annotated (or it wasn't an arrow type): " ++ show e
+
+anns2Class :: Ann.Type -> InferFn Classes.Type
+anns2Class (Ann.Type []) = throwError $ Compiler.TypeError "Cannot change empty Ann type to Classes type"
+anns2Class (Ann.Type ts) = ann2Class $ head ts
+
+ann2Class :: Ann.TypeAnnote -> InferFn Classes.Type
+ann2Class (Ann.TBuiltin bty) = case bty of
```
j-hui, line 189: Note to self: it seems like `Builtin t` is a `Functor` over `t` here, and we should be able to exploit this pattern.. maybe a `mapM` over the `Builtin` type.. (no action needed now but consider for future work.)

src/IR/TypeInference.hs
```diff
@@ -4,72 +4,203 @@ For now, this module implements a very crude form of pure, syntax-directed type
...
+    else throwError $ Compiler.TypeError $ "After expression has inconsistent type: " ++ show e
+inferPrim (I.Prim (I.PrimOp I.PrimSub) [e1, e2]  _) = do
+  e1' <- inferExpr e1
+  e2' <- inferExpr e2
+  return $ I.Prim (I.PrimOp I.PrimSub) [e1', e2'] $ int 32
+inferPrim e@(I.Prim p [] _)
+  | p `elem` [I.Break, I.Return] = return $ I.Prim p [] void
+  | otherwise = throwError $ Compiler.TypeError $ "Unable to type Prim expression: " ++ show e
+inferPrim e@(I.Prim p es _)
+  | p `elem` [I.Loop, I.Wait, I.Par] = do
```
j-hui, line 150: `par` should return a tuple of its argument types

j-hui: Also as a matter of coding style, I prefer an explicit match on each case of primitive, even if there's a bit of repetition.

src/IR/TypeInference.hs
```diff
@@ -4,72 +4,203 @@ For now, this module implements a very crude form of pure, syntax-directed type
...
+      es' <- mapM inferExpr es
+      return $ I.Prim p es' unit
+  | otherwise = throwError $ Compiler.TypeError $ "Unable to type Prim expression: " ++ show e
+inferPrim e = throwError $ Compiler.TypeError $ "Unable to type Prim expression: " ++ show e
+
+localInferExpr :: [(Binder, I.Expr Ann.Type)] -> I.Expr Ann.Type -> InferFn ([(Binder, I.Expr Classes.Type)], I.Expr Classes.Type)
```
j-hui, line 156: Since this helper is only ever used for `let` bindings, perhaps it could be inlined there? Also, the traversal through the list seems like something that could be handled with a `mapM` over a simpler function.

src/IR/TypeInference.hs
```diff
@@ -4,72 +4,203 @@ For now, this module implements a very crude form of pure, syntax-directed type
...
+        I.Var v _ -> insertVar v rrty
+        _ -> return ()
+      return $ I.Prim I.Assign [rrty <$ lhs', rhs'] unit
+    else throwError $ Compiler.TypeError $ "Assign expression has inconsistent type: " ++ show e
+inferPrim e@(I.Prim I.After [del, lhs, rhs] _) = do
+  del' <- inferExpr del
+  rhs' <- inferExpr rhs
+  lhs' <- inferExpr lhs
+  let rrty = ref $ extract rhs'
+  if extract del' == int 32 && extract lhs' == rrty
```
j-hui, line 135: Delay should probably actually be 64-bit int. Can you change that hard-coded int size everywhere?

j-hui (approved):
Awesome work Xijiao, this looks great! I left some suggestions; feel free to take them or leave them as you see fit.

However, before you merge, please add haddock comments to your code, even if it's something brief (e.g., `-- | Infer the type of an expression.`). And also make sure to include module-level documentation about the current capabilities of this compiler pass.

After that's sorted out, feel free to squash & merge

src/IR/TypeInference.hs
XijiaoLi: Yeah we should discuss this more tomorrow!

PR #41, opened 2021-10-27
add 31 bit integer support; blink and hello working

EmilySillars:
# 31 bit integers documentation
Integers are odd 32-bit integers (the final bit of 1 distinguishes them from pointers). The C-translation of SSLANG integer operations are defined below. Most operations use the general translations of binary a OP b  or unary OP a. Translations of addition, subtraction, and bitwise not (~) can be optimized to use fewer steps than the general translations.
The following operations are unaffected by the final bit always being 1:
 - Comparisons Greater than (>), Less than (<), Equal to (==), GTE (>=), LTE (<=)
 - Bitwise AND (&) and btiwise OR (|) 

## a OP b
- a >> 1 //arithmetic    
- b >> 1 //arithmetic
- c = a OP b
- c << 1
- c | 00000001
## OP a
- a >> 1 //arithmetic
- c = OP a
- c << 1
- c | 00000001
## Add/Subtract Optimization
- b & 11111110
- c = a OP b
## 1's Complement Optimization
- c = ~a
- c | 00000001
## a OP b direct translation (for unaffected OPs)
- c = a OP b
# Testing
- currently works with hello and blink
- TODO: make sure edit I made to make blink work won't break other programs
- TODO: make more tests


sedwards-lab:
You can be a little more clever. 1's complement can be expressed x ^ (~1) (Note that ~1 is a constant).

src/Codegen/Codegen.hs
```diff
@@ -521,11 +521,12 @@ genPrim I.After [time, lhs, rhs] _ = do
   (lhsVal , lhsStms ) <- genExpr lhs
   (rhsVal , rhsStms ) <- genExpr rhs
   Just later          <- return $ genLater $ extract lhs
+  let timeUnmarshalled  = unmarshal timeVal --is this a dangerous band aid?
```
j-hui, line 524: What is the issue you're worried about here?

Note that we also haven't actually worked out the types for time units quite yet. For now, I'm assuming all time values are some kind of integer value. If it's the fact that this only works if `timeVal` is some kind of integer value, then don't worry about that. You can assume it will be an integer, and that any kind of type error can and should be caught by the typechecker further upstream.

j-hui: Two stylistic nits: (1) if you ever have consecutive `let` definitions, you can keep them all in the same block, i.e.,
```haskell
let timeUnmarshalled = ...
    laterBlock = ...
```

But the other point is that you should just be able to perform the unmarshalling inline when `timeVal` is bound, using the `first` + `<$>` pattern we discussed before:

```haskell
(timeVal, timeStms) <- first unmarshal <$> genExpr time
```

EmilySillars: Thank you for your explanation! I was worried that the generated C code for later might not always need an unmarshalled timeVal, but I realize it always will.

EmilySillars: Ah, thank you for this note about inlining unmarshalling of timeVal. I didn't notice it was the same pattern but  you're right, it is!

regression-tests/tests/31bitInts.ssl
```diff
@@ -0,0 +1,13 @@
...
+  after 10 , (cout : & Int) <- (9 + 5 + 10 -5 + 90 - 111 + 103):Int
+  wait (cout : & Int)
+  after 10 , (cout : & Int) <- (109 - 1):Int
+  wait (cout : & Int)
+  after 10 , (cout : & Int) <- ((deref (cout:&Int):Int) -86 + 4 + 83 +3 -5 +1):Int
+  wait (cout : & Int)
+  after 10 , (cout : & Int) <- (219 - deref (cout:&Int):Int ):Int
+  wait (cout : & Int)
+  after 10 , (cout : & Int) <- 10
+  wait (cout : & Int)
```
j-hui, line 13: Nit: please add newline at end of file

src/Codegen/Codegen.hs
```diff
@@ -586,72 +586,92 @@ genPrim I.Return [e] _ = do
...
+marshal e = [cexp|((((typename uint32_t) $exp:e) << 1) | 0x1)|]
+
+-- | Unmarshal a C expression evaluating to a SSLANG 31 bit 
+unmarshal :: C.Exp -> C.Exp
+unmarshal e = [cexp|(((typename uint32_t) $exp:e) >> 1)|]
 
 -- | Compute priority and depth arguments for a par fork of width 'numChildren'.
 genParArgs :: Int -> (C.Exp, C.Exp) -> [(C.Exp, C.Exp)]
 genParArgs numChildren (currentPrio, currentDepth) =
+
```
j-hui, line 674: nit: get rid of this newline

src/Codegen/Codegen.hs
```diff
@@ -586,72 +586,92 @@ genPrim I.Return [e] _ = do
 genPrim (I.PrimOp op) es t = genPrimOp op es t
 genPrim _ _ _ = fail "Unsupported Primitive or wrong number of arguments"
 
+
```
j-hui, line 589: nit: get rid of these newlines

src/Codegen/Codegen.hs
```diff
@@ -586,72 +586,92 @@ genPrim I.Return [e] _ = do
...
+genBinop :: Expr -> Expr -> GenFn ((C.Exp, C.Exp), [C.BlockItem])
 genBinop lhs rhs = do
   (lhsVal, lhsStms) <- genExpr lhs
   (rhsVal, rhsStms) <- genExpr rhs
-  return (lhsVal, rhsVal, lhsStms ++ rhsStms)
+  return ((lhsVal, rhsVal), lhsStms ++ rhsStms)
+
+-- | Marshal a C expression evaluating to a 32 bit int
+marshal :: C.Exp -> C.Exp
+marshal e = [cexp|((((typename uint32_t) $exp:e) << 1) | 0x1)|]
```
j-hui, line 665: nit: we use `typename uint32_t` a lot of places, so perhaps this should be extracted into `Codegen.Identifiers`.

Perhaps we could introduce a `word_t`:

```
--- i/src/Codegen/Identifiers.hs
+++ w/src/Codegen/Identifiers.hs
@@ -88,6 +88,9 @@ priority = "priority"
 sv_t :: C.Type
 sv_t = [cty|struct sv|]

+word_t :: C.Type
+word_t = [cty|typename uint32_t|]
+
 -- | The conventional name of the payload field in a type-specialized SV.
 value :: CIdent
 value = "value"
```

and then replace all occurrences of `(typename uint32_t)` with that to make it clearer

src/Codegen/Codegen.hs
```diff
@@ -586,72 +586,92 @@ genPrim I.Return [e] _ = do
...
 genPrimOp I.PrimBitOr [lhs, rhs] _ = do
-  (lhsVal, rhsVal, stms) <- genBinop lhs rhs
+  ((lhsVal, rhsVal), stms) <- genBinop lhs rhs
   return ([cexp|$exp:lhsVal | $exp:rhsVal|], stms)
 genPrimOp I.PrimBitNot [opr] _ = do
   (val, stms) <- genExpr opr
-  return ([cexp|~ $exp:val|], stms)
+  -- all integers are 31 bits + 1 tag bit, so val XOR (~1)
+  -- flips the 31 bits and keeps the tag bit 1. 
+  return ([cexp|($exp:val ^ 0xFFFFFFFE)|], stms)
```
j-hui, line 635: nit: why don't we change `0xFFFFFFFE` to `~1`?

j-hui (approved):
Great work, I left a few style nits but overall LGTM.

Feel free to squash & once you've addressed those nits!

src/Codegen/Codegen.hs
EmilySillars: Since ~1 is a constant, I thought I could save an assembly instruction by writing 0xFFFFFFFE. I can change it to ~1 if it's clearer though.

sedwards-lab: An optimizing compiler will always fold constants like this; the problem with using 0xFFFF.. is that it only works with 32-bit integers and will break on a 64-bit platform, which we should keep in mind.

EmilySillars: I used 0xFFFFFFFE in other parts of the generated C code., like in genPrimOp I.PrimAdd. So I should change all instances of 0xFFFFFFFE  to ~1 to ensure compatibility, right?

PR #42, opened 2021-10-28
Fix builds on macOS

### LINKED ISSUE #35: Ar does not work on macOS
@maxhhelman ran into this issue when we were pair programming. If I remember correctly, `ar` on macOS does not recognize the `-U` flag (and I think Zach ran across this issue before too).

It looks like `-U` is the default behavior anyway: https://www.man7.org/linux/man-pages/man1/ar.1.html

We should adjust the ARFLAGS in the ssm-runtime Makefile, maybe just remove that flag if it's not actually necessary, so that we can support building on macOS.

j-hui:
NOTE: I intend on keeping these commits separate and rebase them, for cleaner upstreaming, so expect force pushes.

Fixes #35 by dropping `-U` flag (see commit message for details).

Still needs a solution to fix macOS's dodgy support for weak symbols.

One solution is to just case over the platform `[ "$(uname)" == "Darwin" ]` and manually specify `-Wl,-U,_ssm_throw` (following advice from https://stackoverflow.com/a/51755953), but this isn't exactly elegant, and we'll have to do this for every weak symbol. I'm looking into an alternate approach that involves weak aliases, which according to http://maskray.me/blog/2021-04-25-weak-symbol is less error prone (perhaps more uniformly supported?) across different toolchains. Will need some help testing this out; @hmontero1205 @XijiaoLi and @kentjhall , do you have suggestions/cycles to help try some things out.

### DIFF
--- a/lib/ssm/Makefile
+++ b/lib/ssm/Makefile
@@ -23,24 +23,20 @@ OBJECTS = $(patsubst src/%.c, build/%.o, $(SOURCES))
 EXAMPLES = $(wildcard examples/*.c)
 EXAMPLEEXES = $(patsubst examples/%.c, build/%, $(EXAMPLES))
 
-RED = \e[31m
-GREEN = \e[32m
-RESET_COLOR = \e[0m
-
 ARFLAGS = -cr
 
 all : test-examples test_main
 
 test_main : build/test_main
-	./build/test_main > build/test_main.out || echo "${RED}TEST_MAIN FAILED${RESET_COLOR}"
+	./build/test_main > build/test_main.out || echo "TEST_MAIN FAILED"
 	@(diff test/test_main.out build/test_main.out && \
-	echo "${GREEN}TEST_MAIN PASSED${RESET_COLOR}") || \
-	echo "${RED}TEST_MAIN OUTPUT DIFFERS${RESET_COLOR}"
+	echo "TEST_MAIN PASSED") || \
+	echo "TEST_MAIN OUTPUT DIFFERS"
 test-examples : examples
 	./runexamples > build/examples.out
 	@(diff test/examples.out build/examples.out && \
-	echo "${GREEN}EXAMPLES PASSED${RESET_COLOR}") || \
-	echo "${RED}EXAMPLE OUTPUT DIFFERS${RESET_COLOR}"
+	echo "EXAMPLES PASSED") || \
+	echo "EXAMPLE OUTPUT DIFFERS"
 
 build/test_main : test/test_main.c build/libssm.a
 	$(CC) $(CFLAGS) -o $@ test/test_main.c -Lbuild -lssm
--- a/lib/ssm/Makefile
+++ b/lib/ssm/Makefile
@@ -64,9 +64,7 @@ build/libssm.a : $(INCLUDES) $(OBJECTS)
 	rm -f build/libssm.a
 	$(AR) $(ARFLAGS) build/libssm.a $(OBJECTS)
 
-$(OBJECTS) : $(INCLUDES) $(SOURCES)
-
-build/%.o : src/%.c
+build/%.o : src/%.c $(INCLUDES)
 	$(CC) $(CFLAGS) -c -o $@ $<
 
 examples : $(EXAMPLEEXES)
--- a/lib/ssm/Makefile
+++ b/lib/ssm/Makefile
@@ -27,7 +27,7 @@ RED = \e[31m
 GREEN = \e[32m
 RESET_COLOR = \e[0m
 
-ARFLAGS = -crU
+ARFLAGS = -cr
 
 all : test-examples test_main
 
--- a/lib/ssm/include/ssm.h
+++ b/lib/ssm/include/ssm.h
@@ -74,8 +74,7 @@ See [the detailed documentation](@ref all)
  * pointer if the linker does not find a definition for this symbol in any
  * object file.
  */
-void ssm_throw(int reason, const char *file, int line, const char *func)
-  __attribute__((weak));
+void ssm_throw(int reason, const char *file, int line, const char *func);
 
 /** Invoked when a process must terminate, e.g., when memory or queue space is
  * exhausted. Not expected to return.
@@ -84,13 +83,7 @@ void ssm_throw(int reason, const char *file, int line, const char *func)
  * Default behavior is to exit with reason as the exit code, but can be
  * overridden by defining ssm_throw.
  */
-#define SSM_THROW(reason) \
-    do \
-      if (ssm_throw) \
-        ssm_throw(reason, __FILE__, __LINE__, __func__); \
-      else \
-        for(;;); \
-    while (0)
+#define SSM_THROW(reason) ssm_throw(reason, __FILE__, __LINE__, __func__)
 
 /** Error codes, indicating reason for failure.
  *
--- a/lib/ssm/src/ssm-scheduler.c
+++ b/lib/ssm/src/ssm-scheduler.c
@@ -297,7 +297,7 @@ void ssm_unschedule(ssm_sv_t *var)
     q_idx_t hole = find_queued_event(var);
     var->later_time = SSM_NEVER;
     ssm_sv_t *moved_var = event_queue[event_queue_len--];
-    if (hole < SSM_QUEUE_HEAD + event_queue_len)
+    if (hole < SSM_QUEUE_HEAD + event_queue_len) {
       // Percolate only if removal led to a hole in the queue; no need to do
       // this if we happened to remove the last element of the queue.
       if (hole == SSM_QUEUE_HEAD ||
@@ -305,6 +305,7 @@ void ssm_unschedule(ssm_sv_t *var)
         event_queue_percolate_down(hole, moved_var);
       else
         event_queue_percolate_up(hole, moved_var);
+    }
   }
 }
 
--- a/lib/ssm/src/ssm-throw.c
+++ b/lib/ssm/src/ssm-throw.c
@@ -1,0 +1,6 @@
+#include <ssm.h>
+
+/** Default, platform-generic implementation of ssm_throw; spins forever. */
+void ssm_throw(int reason, const char *file, int line, const char *func) {
+  for(;;);
+}


sedwards-lab:
What are we using the weak and alias features for, again? We should keep the runtime library code as generic as possible since we're likely to keep running into compatibility problems like this.

j-hui:
The weak symbol was introduced to allow an implementation of `ssm_throw` to be overridden at link time. Our "default" implementation should be just `for (;;)` (rather than `exit()`, which is also not portable and whose behavior is extremely platform-dependent), since that's guaranteed to be platform-generic. However, that isn't very helpful when something _does_ crash, since it will just silently hang.

Using a weak symbol. a platform-specific binding can make sure that a more reasonable implementation is used, e.g.,:

```
ld user_code.o
 + libssm.a // weak definition of ssm_throw
 + libplatform1.a // strong (normal) definition of ssm_throw
=> exe // uses definition given by libplatform1.a
```

whereas if it isn't specified by the platform, it will just default to the generic `for (;;)` implementation:

```
ld user_code.o
 + libssm.a // weak definition of ssm_throw
 + libplatform2.a // no definition of ssm_throw
=> exe // uses definition given by libssm.a
```

In these two scenarios, neither `libssm.a` nor `user_code.o` need to be recompiled.

The alternative is to introduce a preprocessor setting for this, e.g., something like:

```
#ifdef CONFIG_SSM_THROW_OVERRIDE
extern void ssm_throw(int reason, const char *file, int line, const char *func);
#define SSM_THROW(reason) ssm_throw(reason, __FILE__, __LINE__, __FUNC__)
#else
#define SSM_THROW(reason) for(;;)
```

The slight downside is that now we must be careful to compile every compilation unit with the same flags, which means that this is now something we need to expose in the interface of `ssm-runtime`'s build system, and it's not something that `ssm-runtime`'s Makefile can easily check for consistency. For instance, if you accidentally run:
```
$ make build/ssm-bool.o  # compiled with default spin implementation
$ # ... forget to clean up build/ssm-bool.o
$ CONFIG_SSM_THROW_OVERRIDE=1 make build/libssm.a # other object files compiled extern ssm_throw definition
```
Now linking against `libssm.a` is going to lead to some rather unexpected hangs only when those errors emanate from the stale  `ssm-bool.o` compilation unit, which as far as Make is concerned, is up to date.

That being said, using a config may just be our only option, because I don't really see any other platform-generic solution.

j-hui:
Following discussion with @sedwards-lab earlier today, I'm going to remove the weak symbol thing altogether. The symbol can still be overriden at link time by ensuring that the compilation unit defining `ssm_throw` comes before `-lssm` in the arguments passed to `ld`/`cc`.

@hmontero1205 , can you test and see if `make test-throw` works for you, as well as the regression tests?

j-hui:
(btw @sedwards-lab , should I fix that dangling else warning from libssm while I'm at it?)

hmontero1205:
Output for `make`:
```
hans@bigmac:~/columbia/sslang/lib/ssm $ make
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-bool.o src/ssm-bool.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-event.o src/ssm-event.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i16.o src/ssm-i16.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i32.o src/ssm-i32.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i64.o src/ssm-i64.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i8.o src/ssm-i8.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-scheduler.o src/ssm-scheduler.c
src/ssm-scheduler.c:306:7: warning: add explicit braces to avoid dangling else [-Wdangling-else]
      else
      ^
1 warning generated.
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-throw.o src/ssm-throw.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-top-act.o src/ssm-top-act.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u16.o src/ssm-u16.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u32.o src/ssm-u32.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u64.o src/ssm-u64.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u8.o src/ssm-u8.c
rm -f build/libssm.a
ar -cr build/libssm.a  build/ssm-bool.o  build/ssm-event.o  build/ssm-i16.o  build/ssm-i32.o  build/ssm-i64.o  build/ssm-i8.o  build/ssm-scheduler.o  build/ssm-throw.o  build/ssm-top-act.o  build/ssm-u16.o  build/ssm-u32.o  build/ssm-u64.o  build/ssm-u8.o
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/clock examples/clock.c -Lbuild -lssm
examples/clock.c:192:37: warning: format specifies type 'unsigned long' but the argument has type 'unsigned long long' [-Wformat]
  printf("simulated %lu seconds\n", ssm_now() / SSM_SECOND);
                    ~~~             ^~~~~~~~~~~~~~~~~~~~~~
                    %llu
1 warning generated.
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/counter examples/counter.c -Lbuild -lssm
examples/counter.c:356:35: warning: format specifies type 'unsigned long' but the argument has type 'ssm_time_t' (aka 'unsigned long long') [-Wformat]
    printf("finished time %lu\n", ssm_now());
                          ~~~     ^~~~~~~~~
                          %llu
1 warning generated.
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/counter2 examples/counter2.c -Lbuild -lssm
examples/counter2.c:357:35: warning: format specifies type 'unsigned long' but the argument has type 'ssm_time_t' (aka 'unsigned long long') [-Wformat]
    printf("finished time %lu\n", ssm_now());
                          ~~~     ^~~~~~~~~
                          %llu
1 warning generated.
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/fib examples/fib.c -Lbuild -lssm
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/fib2 examples/fib2.c -Lbuild -lssm
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/fib3 examples/fib3.c -Lbuild -lssm
examples/fib3.c:181:37: warning: format specifies type 'unsigned long' but the argument has type 'unsigned long long' [-Wformat]
  printf("simulated %lu seconds\n", ssm_now() / SSM_SECOND);
                    ~~~             ^~~~~~~~~~~~~~~~~~~~~~
                    %llu
1 warning generated.
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/hello examples/hello.c -Lbuild -lssm
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/hello2 examples/hello2.c -Lbuild -lssm
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/onetwo examples/onetwo.c -Lbuild -lssm
examples/onetwo.c:152:35: warning: format specifies type 'unsigned long' but the argument has type 'ssm_time_t' (aka 'unsigned long long') [-Wformat]
    printf("finished time %lu\n", ssm_now());
                          ~~~     ^~~~~~~~~
                          %llu
1 warning generated.
./runexamples > build/examples.out
\e[32mEXAMPLES PASSED\e[0m
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/test_main test/test_main.c -Lbuild -lssm
./build/test_main > build/test_main.out || echo "\e[31mTEST_MAIN FAILED\e[0m"
\e[32mTEST_MAIN PASSED\e[0m
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/test-throw test/test-throw.c -Lbuild -lssm
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/test-throw-override test/test-throw.c test/override-throw.c -Lbuild -lssm
timeout -v 0.01 ./build/test-throw >build/test-throw.out 2>&1 || true
timeout -v 0.01 ./build/test-throw-override >>build/test-throw.out 2>&1 || true
\e[32mEXAMPLES PASSED\e[0m
```

Note that
1. I had to install `timeout` via `brew install coreutils`
2. There's still warnings
3. The colors don't seem to be rendering (are we missing a `-e` flag on echo?)

hmontero1205:
Output for `make test-throw`:
```
hans@bigmac:~/columbia/sslang/lib/ssm $ make test-throw
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-bool.o src/ssm-bool.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-event.o src/ssm-event.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i16.o src/ssm-i16.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i32.o src/ssm-i32.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i64.o src/ssm-i64.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-i8.o src/ssm-i8.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-scheduler.o src/ssm-scheduler.c
src/ssm-scheduler.c:306:7: warning: add explicit braces to avoid dangling else [-Wdangling-else]
      else
      ^
1 warning generated.
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-throw.o src/ssm-throw.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-top-act.o src/ssm-top-act.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u16.o src/ssm-u16.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u32.o src/ssm-u32.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u64.o src/ssm-u64.c
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -c -o build/ssm-u8.o src/ssm-u8.c
rm -f build/libssm.a
ar -cr build/libssm.a  build/ssm-bool.o  build/ssm-event.o  build/ssm-i16.o  build/ssm-i32.o  build/ssm-i64.o  build/ssm-i8.o  build/ssm-scheduler.o  build/ssm-throw.o  build/ssm-top-act.o  build/ssm-u16.o  build/ssm-u32.o  build/ssm-u64.o  build/ssm-u8.o
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/test-throw test/test-throw.c -Lbuild -lssm
cc -Iinclude -O -Wall -pedantic -std=c99 -g -DSSM_DEBUG  -o build/test-throw-override test/test-throw.c test/override-throw.c -Lbuild -lssm
timeout -v 0.01 ./build/test-throw >build/test-throw.out 2>&1 || true
timeout -v 0.01 ./build/test-throw-override >>build/test-throw.out 2>&1 || true
\e[32mEXAMPLES PASSED\e[0m
```

hmontero1205:
fyi the `parsed` tests seem to be failing:
```
hans@bigmac:~/columbia/sslang/regression-tests $ ./runtests.sh
blink-parsed...FAILED
  failed: stack exec sslc -- --module-name=blink-parsed --generate-c tests/blink-parsed.ssl > out/blink-parsed.c
blink-parsed2...FAILED
  failed: stack exec sslc -- --module-name=blink-parsed2 --generate-c tests/blink-parsed2.ssl > out/blink-parsed2.c
blink...OK
doubleblink-parsed...FAILED
  failed: stack exec sslc -- --module-name=doubleblink-parsed --generate-c tests/doubleblink-parsed.ssl > out/doubleblink-parsed.c
doubleblink...OK
hello1...OK
```

Want me to look into this?

hmontero1205:
False alarm, the regression tests are fine:
```
$ ./runtests.sh
blink...OK
doubleblink...OK
hello1...OK
```

Looks like I had some untracked files in the test directory and the suite was trying to test with them.

hmontero1205:
@sedwards-lab 

`echo` is apparently not portable across operating systems (or even shells). On my macOS, `echo` doesn't seem to care about color escape sequences, even when I add a `-e` flag. Stackoverflow suggests using `printf` which is supposed to be more portable:
```
$ git diff
diff --git a/lib/ssm/Makefile b/lib/ssm/Makefile
index 6300a85..aa9ec79 100644
--- a/lib/ssm/Makefile
+++ b/lib/ssm/Makefile
@@ -32,21 +32,21 @@ ARFLAGS = -cr
 all : test-examples test_main test-throw

 test_main : build/test_main
-       ./build/test_main > build/test_main.out || echo "${RED}TEST_MAIN FAILED${RESET_COLOR}"
+       ./build/test_main > build/test_main.out || printf "${RED}TEST_MAIN FAILED${RESET_COLOR}\n"
        @(diff test/test_main.out build/test_main.out && \
-       echo "${GREEN}TEST_MAIN PASSED${RESET_COLOR}") || \
-       echo "${RED}TEST_MAIN OUTPUT DIFFERS${RESET_COLOR}"
+       printf "${GREEN}TEST_MAIN PASSED${RESET_COLOR}\n") || \
+       printf "${RED}TEST_MAIN OUTPUT DIFFERS${RESET_COLOR}\n"
 test-examples : examples
        ./runexamples > build/examples.out
        @(diff test/examples.out build/examples.out && \
-       echo "${GREEN}EXAMPLES PASSED${RESET_COLOR}") || \
-       echo "${RED}EXAMPLE OUTPUT DIFFERS${RESET_COLOR}"
+       printf "${GREEN}EXAMPLES PASSED${RESET_COLOR}\n") || \
+       printf "${RED}EXAMPLE OUTPUT DIFFERS${RESET_COLOR}\n"
 test-throw : build/test-throw build/test-throw-override
        timeout -v 0.01 ./build/test-throw >build/test-throw.out 2>&1 || true
        timeout -v 0.01 ./build/test-throw-override >>build/test-throw.out 2>&1 || true
        @(diff test/test-throw.out build/test-throw.out && \
-       echo "${GREEN}EXAMPLES PASSED${RESET_COLOR}") || \
-       echo "${RED}EXAMPLE OUTPUT DIFFERS${RESET_COLOR}"
+       printf "${GREEN}EXAMPLES PASSED${RESET_COLOR}\n") || \
+       printf "${RED}EXAMPLE OUTPUT DIFFERS${RESET_COLOR}\n"

 build/test_main : test/test_main.c build/libssm.a
        $(CC) $(CFLAGS) -o $@ test/test_main.c -Lbuild -lssm
```

Does that seem like a reasonable change?

j-hui:
> I had to install timeout via brew install coreutils
This is fine---I'll probably remove that test.

I'm not sure how to best work around the formatter warnings. It's almost inevitable without a bit of macro wrangling since the type->size mappings are just different macOS

sedwards-lab:
Colors are a function of the terminal in which you're displaying the results, not the echo command itself. And there are definitely different versions of echo (e.g., some don't pay attention to the -e flag).

My feeling is that we ought to back out all the fancy things (color is fun, but non-portable) and try to make the Makefile as portable as possible

Similarly, the dependence on 'timeout" seems needless. There's a more portable way to terminate long-running processes through the shell.

j-hui:
I agree with removing colors for portability. They're useful for visually identifying failed tests, but we're really not running that many tests. If we dramatically expand this test suite, we can always add colors back into the Makefile.

j-hui:
Apologies for the force pushes. I just rebased everything on top of the latest main; @hmontero1205 can you double check that everything is working for you on macOS?

hmontero1205:
@j-hui everything still works fine. It looks like `test-throw` isn't a Makefile target anymore though, but if that's expected, LGTM.

j-hui:
Yeah, taking out the test-throw target was intentional. I had pushed it
there as a preliminary sanity check for how whether linker behavior differs
between gcc's linker and lld. I replaced that with a more comprehensive
test in ssm-scratch and found that behavior does indeed differ, but not in
a way that matters to us.

sedwards-lab (approved):
Looks good. The colors are fun, but unless there's a way to check whether they're available, better to leave them out. Don't use fancy stuff unless you really need to rely on fancy stuff.

hmontero1205 (approved):
send it!!!!!

PR #43, opened 2021-11-10
Add small fix on type inference and regression test

XijiaoLi:
In this PR:
- Change Int type back to 32-bit
- Update doubleblink2-main.c
- Add `PrimAdd` case to the `inferPrim` function


j-hui (approved):
LGTM

PR #44, opened 2021-11-15
Lift lambdas

### LINKED ISSUE #32: Lambda-lifting
We need to convert lambdas (aka closures) into top-level function definitions. So this:

```
f = \x -> (\() -> x) ()
```

should be converted to

```
type Fenv a = Fenv { x : a }

f__anon1 = \fenv -> \() -> fenv.x

f = \x -> f__anon (Fenv a) ()
```

That is, any lambdas that don't appear at the top-level of a top-level function definition should be hoisted into a name-mangled top-level definition and given an extra "environment" parameter that includes all variables captured by closure.

hmontero1205:
Opening a draft to get some feedback in. Need advice on how to type lifted lambdas and general monadic programming critiques :)

Closes #32

### DIFF
--- a/src/IR/IR.hs
+++ b/src/IR/IR.hs
@@ -192,10 +192,9 @@ collectLambda e = ([], e)
 
 -- | Create a lambda chain given a list of argument-type pairs and a body.
 makeLambdaChain :: TypeSystem t => [(Binder, t)] -> Expr t -> Expr t
-makeLambdaChain [] body = body
-makeLambdaChain ((v, ty) : vs) body =
-  let liftedBody = makeLambdaChain vs body
-  in  Lambda v liftedBody $ arrow ty (extract liftedBody)
+makeLambdaChain []   body = body
+makeLambdaChain args body = foldr chain body args
+  where chain (v, t) b = Lambda v b $ t `arrow` extract b
 
 instance Functor Program where
   fmap f Program { programEntry = e, programDefs = defs, typeDefs = tds } =
--- a/src/IR/IR.hs
+++ b/src/IR/IR.hs
@@ -192,7 +192,6 @@ collectLambda e = ([], e)
 
 -- | Create a lambda chain given a list of argument-type pairs and a body.
 makeLambdaChain :: TypeSystem t => [(Binder, t)] -> Expr t -> Expr t
-makeLambdaChain []   body = body
 makeLambdaChain args body = foldr chain body args
   where chain (v, t) b = Lambda v b $ t `arrow` extract b
 
--- a/src/IR/IR.hs
+++ b/src/IR/IR.hs
@@ -11,6 +11,7 @@ module IR.IR
   , DConId(..)
   , wellFormed
   , collectLambda
+  , makeLambdaChain
   ) where
 import           Common.Identifiers             ( Binder
                                                 , DConId(..)
@@ -20,7 +21,10 @@ import           Common.Identifiers             ( Binder
 
 import           Control.Comonad                ( Comonad(..) )
 import           Data.Bifunctor                 ( Bifunctor(..) )
-import           IR.Types.TypeSystem            ( TypeDef(..) )
+import           IR.Types.TypeSystem            ( TypeDef(..)
+                                                , TypeSystem
+                                                , arrow
+                                                )
 
 import           Common.Pretty
 
@@ -185,6 +189,13 @@ collectLambda :: Expr t -> ([Binder], Expr t)
 collectLambda (Lambda a b _) =
   let (as, body) = collectLambda b in (a : as, body)
 collectLambda e = ([], e)
+
+-- | Create a lambda chain given a list of argument-type pairs and a body.
+makeLambdaChain :: TypeSystem t => [(Binder, t)] -> Expr t -> Expr t
+makeLambdaChain [] body = body
+makeLambdaChain ((v, ty) : vs) body =
+  let liftedBody = makeLambdaChain vs body
+  in  Lambda v liftedBody $ arrow ty (extract liftedBody)
 
 instance Functor Program where
   fmap f Program { programEntry = e, programDefs = defs, typeDefs = tds } =
--- a/src/IR/LambdaLift.hs
+++ b/src/IR/LambdaLift.hs
@@ -16,6 +16,8 @@ import           Control.Monad.State.Lazy       ( MonadState
                                                 , unless
                                                 )
 
+import qualified Data.Bifunctor                as B
+import qualified Data.Map                      as M
 import           Data.Maybe                     ( catMaybes )
 import qualified Data.Set                      as S
 
@@ -23,7 +25,7 @@ import qualified Data.Set                      as S
 data LiftCtx = LiftCtx
   { globalScope  :: S.Set I.VarId
   , currentScope :: S.Set I.VarId
-  , currentFrees :: S.Set I.VarId
+  , freeTypes    :: M.Map I.VarId Poly.Type
   , lifted       :: [(I.VarId, I.Expr Poly.Type)]
   , anonCount    :: Int
   }
@@ -36,12 +38,26 @@ newtype LiftFn a = LiftFn (StateT LiftCtx Compiler.Pass a)
   deriving (MonadError Compiler.Error)  via (StateT LiftCtx Compiler.Pass)
   deriving (MonadState LiftCtx)         via (StateT LiftCtx Compiler.Pass)
 
+exprType :: I.Expr Poly.Type -> Poly.Type
+exprType (I.Lambda _ _ t) = t
+exprType (I.App    _ _ t) = t
+exprType (I.Lit _ t     ) = t
+exprType (I.Var _ t     ) = t
+exprType (I.Prim _ _ t  ) = t
+exprType (I.Let  _ _ t  ) = t
+
+zipArgsWithArrow :: [Binder] -> Poly.Type -> [(Binder, Poly.Type)]
+zipArgsWithArrow (b : bs) (Poly.TBuiltin (Poly.Arrow t ts)) =
+  (b, t) : zipArgsWithArrow bs ts
+zipArgsWithArrow [] _ = []
+zipArgsWithArrow _  _ = error "Expected longer arrow type"
+
 runLiftFn :: LiftFn a -> Compiler.Pass a
 runLiftFn (LiftFn m) = evalStateT
   m
   LiftCtx { globalScope  = S.empty
           , currentScope = S.empty
-          , currentFrees = S.empty
+          , freeTypes    = M.empty
           , lifted       = []
           , anonCount    = 0
           }
@@ -68,45 +84,54 @@ addLifted :: String -> I.Expr Poly.Type -> LiftFn ()
 addLifted name lam =
   modify $ \st -> st { lifted = (I.VarId (Identifier name), lam) : lifted st }
 
-addFreeVar :: I.VarId -> LiftFn ()
-addFreeVar v =
-  modify $ \st -> st { currentFrees = S.insert v $ currentFrees st }
+addFreeVar :: I.VarId -> Poly.Type -> LiftFn ()
+addFreeVar v t = modify $ \st -> st { freeTypes = M.insert v t $ freeTypes st }
 
 newScope :: [I.VarId] -> LiftFn ()
 newScope vs = modify $ \st -> st
   { currentScope = S.union (globalScope st) (S.fromList vs)
-  , currentFrees = S.empty
+  , freeTypes    = M.empty
   }
 
+makeArrow :: Poly.Type -> I.Expr Poly.Type -> Poly.Type
+makeArrow lhsType rhsExpr =
+  Poly.TBuiltin $ Poly.Arrow lhsType (exprType rhsExpr)
+
 makeLiftedLambda
-  :: [I.Binder] -> I.Expr Poly.Type -> Poly.Type -> LiftFn (I.Expr Poly.Type)
-makeLiftedLambda [] body _ = return body
-makeLiftedLambda vs body t = do
-  liftedBody <- makeLiftedLambda (tail vs) body t
-  return (I.Lambda (head vs) liftedBody t)
+  :: [(I.Binder, Poly.Type)] -> I.Expr Poly.Type -> LiftFn (I.Expr Poly.Type)
+makeLiftedLambda []            body = return body
+makeLiftedLambda ((v, t) : vs) body = do
+  liftedBody <- makeLiftedLambda vs body
+  return (I.Lambda v liftedBody $ makeArrow t liftedBody)
 
 liftLambdas'
   :: (I.VarId, I.Expr Poly.Type) -> LiftFn (I.VarId, I.Expr Poly.Type)
 liftLambdas' (v, lam@(I.Lambda _ _ t)) = do
   let (vs, body) = I.collectLambda lam
+      vs'        = zipArgsWithArrow vs t
   newScope $ catMaybes vs
   liftedBody <- liftLambdas body
-  return (v, foldl (\lam' v' -> I.Lambda v' lam' t) liftedBody vs)
+  return
+    ( v
+    , foldl (\lam' (v', t') -> I.Lambda v' lam' $ makeArrow t' lam')
+            liftedBody
+            vs'
+    )
 liftLambdas' _ = error "Expected top-level lambda binding"
 
-descend :: LiftFn a -> LiftFn (a, S.Set VarId)
+descend :: LiftFn a -> LiftFn (a, M.Map VarId Poly.Type)
 descend lb = do
-  savedScope    <- gets currentScope
-  savedFrees    <- gets currentFrees
-  liftedLamBody <- lb
-  lamFrees      <- gets currentFrees
-  modify $ \st -> st { currentScope = savedScope, currentFrees = savedFrees }
-  return (liftedLamBody, lamFrees)
+  savedScope     <- gets currentScope
+  savedFreeTypes <- gets freeTypes
+  liftedLamBody  <- lb
+  lamFreeTypes   <- gets freeTypes
+  modify $ \st -> st { currentScope = savedScope, freeTypes = savedFreeTypes }
+  return (liftedLamBody, lamFreeTypes)
 
 liftLambdas :: I.Expr Poly.Type -> LiftFn (I.Expr Poly.Type)
-liftLambdas n@(I.Var v _) = do
+liftLambdas n@(I.Var v t) = do
   inScope <- inCurrentScope v
-  unless inScope $ addFreeVar v
+  unless inScope $ addFreeVar v t
   return n
 liftLambdas (I.App e1 e2 t) = do
   liftedE1 <- liftLambdas e1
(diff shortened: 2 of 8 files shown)


src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,157 @@
...
+  let (vs, body) = I.collectLambda lam
+  traceM "Lambda"
+  oldCtx <- get
+  newScope $ map (\(Just v) -> v) vs
+  liftedLamBody <- liftLambdas body
+  lamFrees      <- gets currentFrees
+  liftedLam     <- makeLiftedLambda (map Just (S.toList lamFrees) ++ vs)
+                                    liftedLamBody
+                                    t
+  freshNum <- getFresh
```
j-hui, line 124: I would just have `getFresh` return a fresh name automatically, rather than having it return a number then building a string yourself out of it.

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,157 @@
...
+  liftedE1 <- liftLambdas e1
+  liftedE2 <- liftLambdas e2
+  return $ I.App liftedE1 liftedE2 t
+liftLambdas (I.Prim p exprs t) = do
+  liftedExprs <- mapM liftLambdas exprs
+  return $ I.Prim p liftedExprs t
+liftLambdas lam@(I.Lambda _ _ t) = do
+  let (vs, body) = I.collectLambda lam
+  traceM "Lambda"
+  oldCtx <- get
```
j-hui, line 117: Avoid manually saving and restoring state like this. See the `withNewScope` helper that Xijiao uses in the type checker

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,157 @@
...
+  let (vs, body) = I.collectLambda lam
+  newScope $ map (\(Just vi) -> vi) vs
+  liftedBody <- liftLambdas body
+  return (v, foldl (\lam' v' -> I.Lambda v' lam' t) liftedBody vs)
+liftLambdas' _ = error "Expected top-level lambda binding"
+
+liftLambdas :: I.Expr Poly.Type -> LiftFn (I.Expr Poly.Type)
+liftLambdas n@(I.Var v _) = do
+  isNotFree <- inCurrentScope v
+  if isNotFree
```
j-hui, line 102: Why not:

```haskell
when isNotFree $ addFreeVar v
return n
```

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,157 @@
...
+makeLiftedLambda vs body t = do
+  liftedBody <- makeLiftedLambda (tail vs) body t
+  traceM (show vs)
+  return (I.Lambda (head vs) liftedBody t)
+
+liftLambdas'
+  :: (I.VarId, I.Expr Poly.Type) -> LiftFn (I.VarId, I.Expr Poly.Type)
+liftLambdas' (v, lam@(I.Lambda _ _ t)) = do
+  let (vs, body) = I.collectLambda lam
+  newScope $ map (\(Just vi) -> vi) vs
```
j-hui, line 94: that lambda is `fromJust` --- is there a reason why you expect a `Just` here?

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,157 @@
...
+  liftedLam     <- makeLiftedLambda (map Just (S.toList lamFrees) ++ vs)
+                                    liftedLamBody
+                                    t
+  freshNum <- getFresh
+  addLifted ("anon" ++ show freshNum) liftedLam
+  modify $ \st -> st { currentScope = currentScope oldCtx
+                     , currentFrees = currentScope oldCtx
+                     }
+  return
+    (foldl (\app v -> I.App app (I.Var v t) t)
```
j-hui, line 130: Use `$` for the outer application to `return`, and try to remove more layers of parens where possible

j-hui (commented):
Thanks Hans! This is great start. Wrote some style suggestions (:

src/IR/LambdaLift.hs
hmontero1205: If it's not `Just`, then it's a wildcard to my understand. I better explicitly ignore those because this code is going to have a runtime error if it encounters a lambda like that.

hmontero1205:
@XijiaoLi I ran into a very interesting issue while I was trying to make my code pass the regression tests. Consider `main` routine of `doubleblink.ssl`:
```
main(led : &Int) -> () =
  par ((slow: &Int -> ()) (led: &Int): ())
         ((fast: &Int -> ()) (led: &Int): ())
```
While the type annotation says `main` returns a unit type, I don't think that's technically true right? Because the type of the body is `((), ())`. In my code, I have to expand chained lambdas and then collect them which requires me to reconstruct the lambda chain (and therefore type annotate each lambda). I was originally ignoring the original type annotation on the overall function because I assumed it'd be exactly the same as the type annotations in the body after type inference. However, I was wrong because I was failing the regression tests. My code, upon reconstructing the `main` function, incorrectly(?) annotated it as `(&Int) -> ((), ()) which makes codegen fail since we don't have tuples yet.

Is the behavior I encountered expected? Does your type inference unify(?) `((), ())` and `()`?

XijiaoLi:
> @XijiaoLi I ran into a very interesting issue while I was trying to make my code pass the regression tests. Consider `main` routine of `doubleblink.ssl`:
> 
> ```
> main(led : &Int) -> () =
>   par ((slow: &Int -> ()) (led: &Int): ())
>          ((fast: &Int -> ()) (led: &Int): ())
> ```
> 
> While the type annotation says `main` returns a unit type, I don't think that's technically true right? Because the type of the body is `((), ())`. In my code, I have to expand chained lambdas and then collect them which requires me to reconstruct the lambda chain (and therefore type annotate each lambda). I was originally ignoring the original type annotation on the overall function because I assumed it'd be exactly the same as the type annotations in the body after type inference. However, I was wrong because I was failing the regression tests. My code, upon reconstructing the `main` function, incorrectly(?) annotated it as `(&Int) -> ((), ()) which makes codegen fail since we don't have tuples yet.
> 
> Is the behavior I encountered expected? Does your type inference unify(?) `((), ())` and `()`?

It should be `((), ())`. I think it was a mistake that was made in the first version of `doubleblink.ssl` and we just haven't changed it. My type inference should infer the type of `par` as `Tuple [t]` Builtin type. However, I believe that the current Codegen does not support `Tuple` type yet based on [Codegen.hs#L164](https://github.com/ssm-lang/sslang/blob/main/src/Codegen/Codegen.hs#L164).

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,205 @@
...
+exprType (I.Lambda _ _ t) = t
+exprType (I.App    _ _ t) = t
+exprType (I.Lit _ t     ) = t
+exprType (I.Var _ t     ) = t
+exprType (I.Prim _ _ t  ) = t
+exprType (I.Let  _ _ t  ) = t
+exprType (I.Data _ t    ) = t
+exprType (I.Match _ _ t ) = t
+
+zipArgsWithArrow :: [Binder] -> Poly.Type -> [(Binder, Poly.Type)]
```
j-hui, line 51: Can't this be rewritten as:

```haskell
zip bs $ fst collectArrow
```

I recall I had written a `collectArrow :: Type -> ([Type], Type)` that unpacks an arrow type into its arguments and its return type (returns empty list for non-arrow type).

Also try to avoid throwing `error` here. If you need, throw a compiler error using utilities from `Compiler.Pass`.

j-hui (changes requested):
Thanks @hmontero1205 ! This is awesome work. 

This looks good overall. I left a bunch of suggestions inline, mostly places where you could have used existing helpers. 

I'm requesting changes because this is still missing Haddock comments, and a test suite demonstrating correctness. For the test suite, add a new test suite named ir-to-ir where we can write tests that assert the correctness of source-to-source transformations.

hmontero1205:
> It should be `((), ())`. I think it was a mistake that was made in the first version of `doubleblink.ssl` and we just haven't changed it. My type inference should infer the type of `par` as `Tuple [t]` Builtin type (e.g., `((), ())` in this case). However, I believe that the current Codegen does not support `Tuple` type yet based on [Codegen.hs#L164](https://github.com/ssm-lang/sslang/blob/main/src/Codegen/Codegen.hs#L164).

@XijiaoLi I'm confused then -- does your type inference logic compare `main`'s annotated return type of `()` to `((), ())`?

I'm still not sure which type annotation I should be following when I am reconstructing lambdas -- the type annotation on the lambda itself or the type annotation on the lambda bodies. To reiterate, I was doing the latter but I'm not sure if I am supposed to be doing the former or if it really shouldn't matter to this compiler pass? @j-hui any thoughts?

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,205 @@
...
+import qualified Data.Bifunctor                as B
+import qualified Data.Foldable                 as F
+import qualified Data.Map                      as M
+import           Data.Maybe                     ( catMaybes )
+import qualified Data.Set                      as S
+
+data LiftCtx = LiftCtx
+  { globalScope  :: S.Set I.VarId
+  , currentScope :: S.Set I.VarId
+  , freeTypes    :: M.Map I.VarId Poly.Type
```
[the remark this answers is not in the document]
hmontero1205: Because I don't need their types. Do you think making them all maps would be more extensible?

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,205 @@
...
+    liftedLamBody
+  freshName <- getFresh
+  addLifted freshName liftedLam
+  return $ applyFreesToLambda
+    (I.Var (I.VarId (Identifier freshName)) (exprType liftedLam))
+    (M.toList lamFreeTypes)
+    (exprType liftedLam)
+ where
+  applyFreesToLambda app ((v', t') : vs) (Poly.TBuiltin (Poly.Arrow _ tr)) =
+    -- TODO(hans): We could assert t' == tl
```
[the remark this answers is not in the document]
hmontero1205: Yeah, it was just a thought. If they aren't equal then something must've gone wrong in type checking.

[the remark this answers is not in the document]
hmontero1205: / type inference

src/IR/LambdaLift.hs
```diff
@@ -0,0 +1,205 @@
...
+      exprs = map snd bs
+  mapM_ addCurrentScope (catMaybes vs)
+  liftedLetBodies <- mapM liftLambdas exprs
+  liftedExpr      <- liftLambdas e
+  return $ I.Let (zip vs liftedLetBodies) liftedExpr t
+liftLambdas (I.Match s arms t) = do
+  liftedMatch <- liftLambdas s
+  liftedArms  <- mapM liftLambdasInArm arms
+  return $ I.Match liftedMatch liftedArms t
+liftLambdas n = return n
```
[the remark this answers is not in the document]
hmontero1205: I support all AST nodes that have exprs in them. This "match-all" is meant for things like literals that don't have exprs inside of them that I need to traverse. Should I be explicit?

j-hui:
`par` should return a tuple of all its expressions.

> I'm still not sure which type annotation I should be following when I am
reconstructing lambdas -- the type annotation on the lambda itself or the
type annotation on the lambda bodies

The two should be more or less equivalent; I think the difference you're
seeing is not intentional. I think it's easier to use the type annotation
on the overall lambda.

src/IR/LambdaLift.hs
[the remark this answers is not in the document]
j-hui: Ah gotcha. You should be more explicit---at least indicate what you've said in an accompanying comment.

src/IR/LambdaLift.hs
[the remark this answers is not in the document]
j-hui: What is `freeTypes`?

XijiaoLi:
> @XijiaoLi I'm confused then -- does your type inference logic compare `main`'s annotated return type of `()` to `((), ())`?
> 
> I'm still not sure which type annotation I should be following when I am reconstructing lambdas -- the type annotation on the lambda itself or the type annotation on the lambda bodies. To reiterate, I was doing the latter but I'm not sure if I am supposed to be doing the former or if it really shouldn't matter to this compiler pass? @j-hui any thoughts?

I mean, the current type annotation for `main`'s return type is wrong - it should be `((), ())` since the type of `par` should be a tuple. The type annotation on the lambda bodies should always be the same as the type annotation on the lambda's return type. We should really fix this regression test.

j-hui:
@XijiaoLi could you issue a PR to fix that? Happy to stamp it.

src/IR/LambdaLift.hs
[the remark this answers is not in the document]
hmontero1205: Ah sorry I didn't respond to that. Yes, it maps ID to type.

src/IR/LambdaLift.hs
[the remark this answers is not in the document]
hmontero1205: I'll remove this comment, this should already be asserted by type inference. I wrote the comment because logically that expression should hold true.

src/IR/LambdaLift.hs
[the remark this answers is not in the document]
j-hui: Doesn't that mean you need their types, if you're maintaining this separate map?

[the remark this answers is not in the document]
hmontero1205: No, I maintain this map because I need the types of the free variables to construct properly typed lambdas

I simply need a set of the names in the current scope so I know when something is/isn't a free variable. I don't care about the types of names in the current scope. Likewise, I carry around the global scope so I can properly descend into new scopes, where the only thing that should be in the new scope is the global scope and possibly some new names.

test/ir-to-ir/Tests/LiftProgramLambdasSpec.hs
```diff
@@ -0,0 +1,84 @@
...
+          , "foo y: Int -> Int ="
+          , "  let w = 1"
+          , "      adder (z: Int) -> Int = z + bar + w"
+          , "  adder y"
+          ]
+        liftedProgram = parseProgram $ unlines
+          [ "bar: Int = 5"
+          , "baz x: Int -> Int ="
+          , "  x + 1"
+          , "anon0 (w: Int) (z: Int) -> Int ="
```
hmontero1205, line 75: fyi @j-hui as I was writing this, I discovered that the compiler doesn't like `anon0 w z: Int -> Int -> Int`. Any idea why? I get this error:

`TypeError "Cannot change empty Ann type to Classes type"`

Should I open a separate issue?

hmontero1205 (commented):
@j-hui check this out, now with tests and documentation 😳

test/ir-to-ir/Tests/LiftProgramLambdasSpec.hs
XijiaoLi: Interesting. This error is generated inside type inference and is caused by no/incomplete type annotation of lambda functions. I tried to print out the lowered AST of this example to see how the type annotations look like at that stage (i.e., right before type inference), and I got:
```
    ( anon0
    , Lambda
      (Just x)
      (Lambda
        (Just z)
        (Prim
          (PrimOp PrimAdd)
          [ Prim (PrimOp PrimAdd) [Var x (Type []), Var z (Type [])] (Type [])
          , Lit (LitIntegral 1) (Type [])
          ]
          (Type [])
        )
        (Type [TBuiltin (Arrow (Type []) (Type []))])
      )
      (Type
        [ TBuiltin
          (Arrow
            (Type [TBuiltin (Integral 32)])
            (Type
              [ TBuiltin
                  (Arrow (Type [TBuiltin (Integral 32)])
                         (Type [TBuiltin (Integral 32)])
                  )
              ]
            )
          )
        , TBuiltin
          (Arrow (Type []) (Type [TBuiltin (Arrow (Type []) (Type []))]))
        ]
      )
    )
```
It seems like even though we have annotated the types of both inputs, somehow it get lost. Currently the type inference on main branch still requires that the input of lambda functions has its type annotated, so thus we got this error.

hmontero1205: That's weird because `baz x: Int -> Int` works fine.

image

XijiaoLi: Ah I think the problem is inside `lowerProgram`, we failed to pass along the type information when we recursively `lowerLambda`. Can you please try to change the `id` in [this line](https://github.com/ssm-lang/sslang/blob/main/src/IR/LowerAst.hs#L165) to `k` and see if it works? I honestly don't really know what's going on inside `LowerAST` LOL.

j-hui: I don't think that line is the issue (and I don't see what we could change it to either).

The information isn't being lost either; it's present in the first type annotation of the entire lambda term, `Int -> Int -> Int`. That should be unified with the second, less specific annotation `_ -> _ -> _`. 

We need to obtain a trace to figure out exactly what IR term is causing this error.

XijiaoLi: Sorry, you are right -- I misunderstood the `k` here. Changing `id` to `k` will simply append `Int -> Int -> Int` to the type annotation of inner lambda (`z -> z + bar + w`), which is incorrect. My bad.

src/IR/LambdaLift.hs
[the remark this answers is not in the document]
j-hui: Could you document what each of these members are in the code?

src/IR/IR.hs
```diff
@@ -186,6 +190,13 @@ collectLambda (Lambda a b _) =
   let (as, body) = collectLambda b in (a : as, body)
 collectLambda e = ([], e)
 
+-- | Create a lambda chain given a list of argument-type pairs and a body.
+makeLambdaChain :: TypeSystem t => [(Binder, t)] -> Expr t -> Expr t
+makeLambdaChain [] body = body
```
j-hui, line 195: why not:

```haskell
makeLambdaChain args body = foldr chain body args
  where chain (v, t) b = Lambda v b $ t `arrow` extract b
```

test/ir-to-ir/Tests/LiftProgramLambdasSpec.hs
```diff
@@ -0,0 +1,84 @@
...
+          , "  x + 1"
+          , "anon0 z: Int -> Int ="
+          , "  z + 1"
+          , "foo y: Int -> Int ="
+          , "  let adder = anon0"
+          , "  adder y"
+          ]
+        nestedToLifted = lowerAndLift nonLiftedProgram
+        liftedToLifted = lowerAndLift liftedProgram
+    nestedToLifted `shouldBe` liftedToLifted
```
j-hui, line 59: Suggestion, though this might be more effort than it's worth: when we compare the two programs, we don't really want to compare the mangled names. Not a blocker for this PR, but it suggests that we should have some kind of more sophisticated IR comparison mechanisms modulo alpha renaming.

test/ir-to-ir/Tests/LiftProgramLambdasSpec.hs
j-hui: Was this ever resolved?

> We need to obtain a trace to figure out exactly what IR term is causing this error.

Did we ever figure out what was causing the issue?

j-hui (approved):
LGTM overall! I'm going to stamp this for now, but there still are a few things to follow up on, for which I left comments inline.

Also, could we add a more deeply nested example? Something like:

```
f x y =
  let z = 4
  let g a =
    let h b = a + b + x
    h z
  g y
```

src/IR/IR.hs
hmontero1205: nice

src/IR/IR.hs
```diff
@@ -186,6 +190,12 @@ collectLambda (Lambda a b _) =
   let (as, body) = collectLambda b in (a : as, body)
 collectLambda e = ([], e)
 
+-- | Create a lambda chain given a list of argument-type pairs and a body.
+makeLambdaChain :: TypeSystem t => [(Binder, t)] -> Expr t -> Expr t
+makeLambdaChain []   body = body
```
j-hui, line 195: You don't need this base case anymore; `foldr` handles that for you

hmontero1205: oops, forgot to remove that, ty!

test/ir-to-ir/Tests/LiftProgramLambdasSpec.hs
hmontero1205: Opened issue #61

test/ir-to-ir/Tests/LiftProgramLambdasSpec.hs
hmontero1205: Opened #62

PR #45, opened 2021-11-15
CI: use GitHub Actions for running tests on PR / Push events

XijiaoLi:
In this PR:
- Added basic Github Actions CI workflow config
  - os: ubuntu-18.04
  - resolver: lts-18.6

### DIFF
--- a/.github/workflows/tester.yml
+++ b/.github/workflows/tester.yml
@@ -1,0 +1,64 @@
+name: Tester
+
+on:
+  push:
+    paths:
+      - '.github/**'
+      - 'app/**'
+      - 'lib/**'
+      - 'regression-tests/**'
+      - 'src/**'
+      - 'test/**'
+      - 'unit-tests/**'
+  pull_request:
+    paths:
+      - '.github/**'
+      - 'app/**'
+      - 'lib/**'
+      - 'regression-tests/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/tester.yml
```diff
@@ -0,0 +1,64 @@
...
+      - 'test/**'
+      - 'unit-tests/**'
+
+jobs:
+  test:
+    name: Build and test sslang
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
```
j-hui, line 29: Is there a reason this uses 18.04? Why not 20.04 (or whatever the latest LTS is)?

j-hui (approved):
Thanks for setting this up Xijiao! This is awesome (: 

Will approve now, but could you try to set this up for Ubuntu 20.04 before merging?

.github/workflows/tester.yml
XijiaoLi: No particular reason; I just chose it randomly xD. Should we change it to `ubuntu-latest` or `ubuntu-20.04` (explicitly fix the version)?

j-hui:
Let's use ubuntu-latest.

PR #46, opened 2021-11-15
Adt c struct gen

EmilySillars:
- I edited typegen.hs so it generates C struct definitions
- This code assumes all built-in types are a word. This should be changed later to account for characters, right?
- I included a struct definition specific to each ADT, but I don't think we actually need it since initializing the struct does not require a name
- My code is based on this example
```
// given these are already defined:
#define ssm_int_obj(A) (((A) << 1) | 0x1)

struct ssm_object_t* ssm_new(uint32_t size, uint8_t tag);

struct ssm_mm_md{
	uint8_t tag;
	uint8_t refcount;
	uint8_t size; // should all be uint8_t which is always 8 bits
};

struct ssm_object_t;

struct ssm_object_t{               // a generic object
	struct ssm_mm_md header;       // gc header
	struct ssm_object_t* payload[1];  // actual length of payload is in header
};

typedef union{
	struct ssm_object_t * heapObj; // points to a generic object
	uint32_t intVal;               // this is our odd 31 bit integer value
}ssm_value_t;

/* Example 1 ****************************************************************/
// data MyBool = MyTrue int | MyFalse int int 

enum myBoolTag {MyTrueTag = 0, MyFalseTag};

struct MyBool{
	struct ssm_mm_md header;
	struct ssm_object_t* payload[2]; //max number of fields	
};

// let a = MyFalse 42 18 

void main(){
	// (max size across all data constructors, tag)
	struct ssm_object_t* a = ssm_new(2, MyFalseTag);
	a->payload[0] = (struct ssm_object_t*) ssm_int_obj(42);
	a->payload[1] = (struct ssm_object_t*) ssm_int_obj(18);
}
```


src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+
+  -- | find any data constructors that hold less than 32 bits of data
+  (big, small ) = partition (\(_, sz) -> sz >= 32) $ augmentWSize <$> dCons
+  -- | find any data constructors that are < 32 bits when including their tag bits
+  (_  , medium) = smallEnough small
+  -- | arrange data constructors from smallest to largest tag value
+  tagged        = small ++ medium ++ big
+
+  -- | define parts of the c definitions
+  enumName      = CIdent $ fromString $ lower $ ident tconid
```
j-hui, line 44: We shouldn't need `lower` here; we should avoid name mangling at this late stage of the compiler. In any case, it's totally fine if the generated code does not follow typical style conventions for identifier names.

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+                       $ty:ssm_object_t* $id:payload[($ty:word_t ) $uint:maxNumFields]; 
+                     } $id:tconid;     
+                   |]
+  tags = [cedecl| enum $id:enumName {
+                          $enums:(enumElts)
+                       };  
+             |]
+
+  -- | find any data constructors that hold less than 32 bits of data
+  (big, small ) = partition (\(_, sz) -> sz >= 32) $ augmentWSize <$> dCons
```
j-hui, line 37: You should be able to replace `(\(_, sz) -> sz >= 32)` with `(>= 32)`

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+  -- | arrange data constructors from smallest to largest tag value
+  tagged        = small ++ medium ++ big
+
+  -- | define parts of the c definitions
+  enumName      = CIdent $ fromString $ lower $ ident tconid
+  enumElts =
+    [cenum|$id:(fst $ fst $ head tagged)|]
+      : (toCEnum . fst . fst <$> tail tagged)
+  maxNumFields =
+    snd $ maximumBy (\(_, sz1) (_, sz2) -> sz1 `compare` sz2) (big ++ medium)
```
j-hui, line 49: You can write ```(\(_, sz1) (_, sz2) -> sz1 `compare` sz2)``` as ```(compare `on` snd)``` (see https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Function.html#v:on)

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+
+  -- | helper functions
```
j-hui, line 51: Remove this line

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+                                                )
+
 -- | Generate definitions for SSM type definitions.
-genTypeDef :: L.TypeDef L.Type -> C.Definition
-genTypeDef = error "Not yet implemented"
+genTypeDef :: TConId -> L.TypeDef L.Type -> (C.Definition, C.Definition)
+genTypeDef tconid (L.TypeDef dCons _) = (tags, structDef)
+ where
+  structDef = [cedecl| typedef struct {
+                       $ty:ssm_mm_md* $id:header;
```
j-hui, line 27: Both the header field and the payload field should be values rather than pointers.

src/Codegen/TypeDef.hs
j-hui: In fact, why not just drop `augmentWSize` altogether and write:

```haskell
partition ((>= 32) . dConSize) dCons
```
You can always recompute `dConSize` given any `TypeVariant`.

src/Codegen/TypeDef.hs
j-hui: Actually, scratch that. It would be better if you didn't pack everything into tuples in the first place, so you could just do ```compare `on` dConSize```

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+  toCEnum n = [cenum|$id:n|]
+
+  -- | categorize data constructor as either small enough or too big to be represented in 31 bits
+  -- | when taking into account the tag bits
+  smallEnough
+    :: [((DConId, L.TypeVariant L.Type), Int)]
+    -> ( [((DConId, L.TypeVariant L.Type), Int)]
+       , [((DConId, L.TypeVariant L.Type), Int)]
+       )
+  smallEnough candidates = partition (\(_, sz) -> (sz + tagBits) < 32)
```
j-hui, line 64: this can be rewritten as ```partition ((< 32) . (+ tagBits) . dConSize) candidates```

j-hui: better yet, introuce a named helper under `tagBits` to make it clear what you're partition on.

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,87 @@
...
+                                     candidates
+    where tagBits = q + r where (q, r) = quotRem (length candidates) 2
```
j-hui, line 66: I think conventional haskell style would write this as: ```length candidates `quotRem` 2```

j-hui (changes requested):
Thanks @EmilySillars ! This is a great start.

I didn't have time to fully work through all the details, but I get the gist of what you're doing. I think it could be a little cleaner though. In particular, instead of maintaining everything as a list of tuples of (variant, size), you can always recover the size by calling `dConSize`. So it might be easier to do everything like that instead of constructing a bunch of tuples, and then constantly destructing them everywhere to get at the size again.

I also left a bunch of other suggestions inline.

A couple more things about style:

- For haddock doc comments, only the first line needs to begin with `|`. See [Kowainik](https://kowainik.github.io/posts/2019-02-06-style-guide#comments) for details about style.
- Also, if you're using any Haskell literals in the doc comment, link them using `'`. For instance, `'DConId`. In the generated documentation, those will show up as clickable links.
- The definitions in the `where` block jump up and down in terms of the order they refer to one another. Try to ensure that definitions read as linearly as possible, i.e., definitions further down should not refer to definitions further up, or vice versa.

src/Codegen/TypeDef.hs
EmilySillars: Ah, yes, I will fix that header to be a value! 
I have a question about the payload being a value - I tried to do this, but the C compiler won't allow it because of cyclic association. In our meeting last week we came up with a struct definition like this:
```
struct ssm_object_t{               // a generic object
	struct ssm_mm_md header;       // gc header
	struct ssm_object_t payload[1];  // causes compiler error! 
};
```
We get a compiler error here because ssm_object_t cannot hold an array of itself/size of the array would be unknown. Because of this error, I changed payload to an array of ssm_object_t pointers like this:
```
struct ssm_object_t{               // a generic object
	struct ssm_mm_md header;       // gc header
	struct ssm_object_t* payload[1];  // compiles okay
};
```
Is there a better way to solve this issue? I think the root of this issue is the initialization code we came up with at the meeting, which never actually use our adt-specific struct, only ssm_object_t:
```
void main(){
	// (max size across all data constructors, tag)
	struct ssm_object_t* a = ssm_new(2, MyFalseTag);
	a->payload[0] = (struct ssm_object_t*) ssm_int_obj(42);
	a->payload[1] = (struct ssm_object_t*) ssm_int_obj(18);
}
```
Is the better way to solve this cyclic association problem to use the adt-specific struct somehow instead of the ssm_object_t struct in this initialization code? Maybe like this??
```
struct MyBool{
	struct ssm_mm_md header;
	struct ssm_object_t payload[2]; //max number of fields	
};
void main(){
	// (max size across all data constructors, tag)
	struct MyBool a = ssm_new(2, MyFalseTag);
	a.payload[0] = ssm_int_obj(42);
	a.payload[1] = ssm_int_obj(18);
}
```

sedwards-lab: There are two types: the single-word value or pointer:

```
union ssm_object_t {
  int value;
  struct ssm_heap_object_t *object;
};
```

and the on-heap object:

```
struct ssm_heap_object_t {
 struct ssm_smm_md header;
 struct ssm_object_t payload[1];
};
```

This should fix the apparent cyclic type dependency because the C compiler doesn't need to know the size of the struct ssm_heap_object_t when it's allocating a pointer for it.

Although it's a little redundant, it probably makes sense to introduce a new struct type for each user-defined ADT, e.g.,

```
struct MyBool {
  struct ssm_mm_md header;
  struct ssm_object_t payload[2]; // max number of fields
};
```

It's redundant because every ADT with two fields max will look like this, but that's not terrible The advantage here is that we can generate code like `sizeof(struct MyBool)` and let the C compiler compute the right number of bytes.

Furthermore, we can "safely" generate code like myvar.object->payload[5] even though the object field is only an array of size 1 in the original type.

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,121 @@
...
+    tagBits = q + r where (q, r) = length candidates `quotRem` 2
+
+  -- | Idenfity int objects by considering the space needed for tag bits
+  (moreHeapObs, intObjs) = smallEnough possibleIntObjs
+
+  -- | Arrange data constructors from smallest to largest tag value
+  tagged                 = intObjs ++ moreHeapObs ++ heapObjs
+
+  -- | Define enum name as 'TConId ++ Tag 
+  enumName               = CIdent $ fromString $ ident tconid ++ "Tag"
```
j-hui, line 80: No need to append `++ "Tag"` to the identifier; we'll handle name mangling elsewhere

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,121 @@
...
+  -- | Save corresponding 'TCon for each 'DCon
+  dConTyp =  M.fromList $ zip (ident.fst <$> dCons) (repeat (ident tconid))
+  -- | Save the size of type 'TCon
+  typSize = M.fromList [(ident tconid,maxNumFields)]
+  -- | Save whether each 'DCon is a pointer or integer
+  isPtr = M.fromList $ theInts ++ thePtrs
+  -- | Save tag of each 'DCon
+  tagMap =  M.fromList $ zip (ident.fst <$> tagged) [0::Int,1..]
+  -- | Save compelete lookup table
+  info = TypeDefInfo dConTyp typSize isPtr tagMap
```
j-hui, line 119: Use named fields for constructing a Record

j-hui (commented):
Feedback given in person, but here are some other straggling notes I took

Also, run everything through Brittany

src/Codegen/TypeDef.hs
EmilySillars: C doesn't let an enum and struct share the same identifier. If I have `data MyBool = MyTrue int | MyFalse int int `, once the C struct is named MyBool, I can't use MyBool again to name the enum. Should I ditch the ADT-specific C struct and only use generic ADT structs in initialization? This way the "MyBool" name would be available to use with the enum.

src/Codegen/TypeDef.hs
```diff
@@ -1,10 +1,121 @@
...
+
```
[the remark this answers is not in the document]
EmilySillars: I thought all the files were supposed to end with a newline - is this different from having a trailing empty line?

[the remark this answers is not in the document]
j-hui: Yeah. The fact that there's an _empty_ line shown on Github's interface means that there's two trailing newlines here.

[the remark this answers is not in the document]
EmilySillars: Ah, thank you, I will fix this!

j-hui:
Superceded by #60

PR #47, opened 2021-11-15
Use shouldBe to replace print in type inference unit test

XijiaoLi:


j-hui (approved):
thanks @XijiaoLi ! LGTM

PR #48, opened 2021-11-15
Added pattern matches to LRM

maxhhelman:


doc/sslang-lrm.tex
```diff
@@ -525,6 +525,7 @@ \section{Expressions}
 \alt "if" <expr_blk> "{" <expr> "}" <expr_else>$^{?}$
 \alt "while" <expr_blk> "{" <expr> "}"
 \alt "fun" <pat>$^{+}$ "{" expr "}"
+\alt "match" <exprBlk> "{"< matchArms> "}"
```
hmontero1205, line 528: 1. use consistent casing: `expr_blk`, `match_arms`.
2. you're missing the grammar for `match_arms`
3. honestly dont make a separate grammar for `match_arms`, look what I did for `wait` and `par`. Use regex syntax.

doc/sslang-lrm.tex
```diff
@@ -715,7 +716,26 @@ \subsection{Sequences}
 
 
 \subsection{Pattern Matching}
-\fixme{todo}
+\begin{grammar}
+<exprBlk> ::= "match" <exprBlk> "{" <matchArms> "}"
+
+<matchArms> ::= matchArm "|" matchArms
```
hmontero1205, line 722: 1. see comments above about casing and folding the `match_arms` inline
2. do something like `<match> := ...` to be consistent with other sections.

doc/sslang-lrm.tex
```diff
@@ -715,7 +716,26 @@ \subsection{Sequences}
...
+
+\begin{lstlisting}
+y = 0
+
+match x // Match clause
+    (Some a)  => y <- 1 // Arm if x is Some a
+    None      => y <- 2 // Arm if x is None
+\end{lstlisting}
+
+This expression checks if the scrutinee \texttt{x} is \texttt{Some a}; if it is, \texttt{y} is set to \texttt{1}. Otherwise, if \texttt{x} is \texttt{None},\texttt{y} is set to \texttt{2}. Pattern matches can also be expressed as \texttt{if-else} statements. However, pattern matches are a much more elegant way to handle multiple cases, especially when we are matching on types.
```
hmontero1205, line 738: I'd move this paragraph before the code listing and start off with "The below code sample checks...."

hmontero1205, line 738: > Pattern matches can also be expressed as \texttt{if-else} statements.

I'm not sure if this is actually true. Pattern matches let you destruct complex ADTs. Take tuples for example.. I think traditionally you only have `fst` and `snd`. You need some pattern-match to break it down.

I'd replace the final two sentences with a brief discussion on how pattern matches are used to destruct ADTs.

hmontero1205 (changes requested):

doc/sslang-lrm.tex
maxhhelman: I have changed the casing. I did include the grammar for match_arms: <match_arms> ::= match_arm "|" match_arms
\alt match_arm

doc/sslang-lrm.tex
maxhhelman: Done

doc/sslang-lrm.tex
maxhhelman: Done

doc/sslang-lrm.tex
```diff
@@ -525,6 +525,7 @@ \section{Expressions}
 \alt "if" <expr_blk> "{" <expr> "}" <expr_else>$^{?}$
 \alt "while" <expr_blk> "{" <expr> "}"
 \alt "fun" <pat>$^{+}$ "{" expr "}"
+\alt "match" <expr_blk> "{"< match_arms> "}"
```
hmontero1205, line 528: In this grammar section, you are missing the definition for `match_arms`. I think you missed one of my comments from last night. Notice what I did for `wait` and `par`. I folded the definition of `par_arms` directly into `par` and `wait` using regex notation.

doc/sslang-lrm.tex
```diff
@@ -715,7 +716,26 @@ \subsection{Sequences}
 
 
 \subsection{Pattern Matching}
-\fixme{todo}
+\begin{grammar}
+<match> ::= "match" <expr_blk> "{" <match_arms> "}"
+
+<match_arms> ::= match_arm "|" match_arms
```
hmontero1205, line 722: see comment above about folding `match_arms` into `match`.

hmontero1205: you also shouldn't need `match_arm`.

hmontero1205 (changes requested):

doc/sslang-lrm.tex
maxhhelman: Done

doc/sslang-lrm.tex
```diff
@@ -715,7 +716,21 @@ \subsection{Sequences}
...
+
+Pattern matches are a succinct way to handle of a variety of cases for the expression in the match clause, known as the scrutinee. If the scrutinee satisfies the pattern in a match arm, the expression in the match arm is evaluated; otherwise, the expression in that arm is not evaluated.
+
+The below code sample checks if the scrutinee \texttt{x} is \texttt{Some a}; if it is, \texttt{y} is set to \texttt{1}. Otherwise, if \texttt{x} is \texttt{None},\texttt{y} is set to \texttt{2}. Pattern matches are an elegant way to handle multiple cases of \texttt{if} statements. They also allow for the destruction of complex algebraic data types via comparison to the patterns in the match arms.
+
+\begin{lstlisting}
+y = 0
+
+match x // Match clause
+    (Some a)  => y <- 1 // Arm if x is Some a
```
j-hui, line 731: Parentheses not necessary, but already addressed in my suggestion for example

doc/sslang-lrm.tex
```diff
@@ -715,7 +716,21 @@ \subsection{Sequences}
 
 
 \subsection{Pattern Matching}
-\fixme{todo}
+\begin{grammar}
+<match> ::=  "match" <expr_blk> "{" (  <pat> "=>" "{" <expr> "}" )$^{+}$ "}"
+\end{grammar}
+
+Pattern matches are a succinct way to handle of a variety of cases for the expression in the match clause, known as the scrutinee. If the scrutinee satisfies the pattern in a match arm, the expression in the match arm is evaluated; otherwise, the expression in that arm is not evaluated.
```
j-hui, line 723: Saying "succinct way" implies that there are other ways/primitives to accomplish the same task; there aren't. This also suggests that if there are multiple matching arms, they will all be evaluated, which isn't true either.

Let's write:

> Pattern matches are used to break down algebraic data types. A pattern match consists of a ``scrutinee''---what is being matched on---and a number of ``arms.'' Each arm consists of a pattern that the scrutinee is matched against, and a ``consequence'' expression. The value of the entire pattern match is the consequence of the first matching arm. The types of each consequence must be the same in a pattern match, and the type of the scrutinee must match the type of the each pattern; the overall type of the match expression is that of its consequences.

doc/sslang-lrm.tex
```diff
@@ -715,7 +716,21 @@ \subsection{Sequences}
...
+
+The below code sample checks if the scrutinee \texttt{x} is \texttt{Some a}; if it is, \texttt{y} is set to \texttt{1}. Otherwise, if \texttt{x} is \texttt{None},\texttt{y} is set to \texttt{2}. Pattern matches are an elegant way to handle multiple cases of \texttt{if} statements. They also allow for the destruction of complex algebraic data types via comparison to the patterns in the match arms.
+
+\begin{lstlisting}
```
j-hui, line 727: I would take out the comments from this listing; they don't really add very much other than narrating the syntax. `y = 0` is also not valid grammar (you need a leading `let`), and even if it were, `y` cannot be reassigned.

How about just this:

```
match x
  Some t => t
  None => 0
 ```

This sample code scrutinizes `x`, an `Option Int`. It unwraps the `Option` type into a plain `Int`, and collapses `None` values into `0`.

**Review by j-hui — CHANGES_REQUESTED** [review]
@maxhhelman sorry for taking so long to get to this. I have some suggestions about the prose; lemme know if they make sense?

@hmontero1205  did you have any other feedback?

**Comment by hmontero1205** [comment]
I'm good for now

**Comment by j-hui** [comment]
@maxhhelman would you like me to take over on this PR?

**Comment by maxhhelman** [comment]
Yeah sounds good

File: doc/sslang-lrm.tex [hunk 9]
```diff
@@ -717,19 +717,27 @@ \subsection{Sequences}
 
 \subsection{Pattern Matching}
 \begin{grammar}
-<match> ::=  "match" <expr_blk> "{" (  <pat> "=>" "{" <expr> "}" )$^{+}$ "}"
+<match> ::=  "match" <expr_blk> "{" (  <pat> "=" "{" <expr> "}" )$^{+}$ "}"
 \end{grammar}
 
-Pattern matches are a succinct way to handle of a variety of cases for the expression in the match clause, known as the scrutinee. If the scrutinee satisfies the pattern in a match arm, the expression in the match arm is evaluated; otherwise, the expression in that arm is not evaluated.
+Pattern matches are used to break down algebraic data types.
+A pattern match consists of a ``scrutinee''---what is being matched on---and a number of ``arms.''
```
hmontero1205, line 724: nit: . outside of "

hmontero1205 (approved):
pls just fix the super minor nit

doc/sslang-lrm.tex
j-hui: Some will say that periods are supposed to be inside of sentences: https://www.grammar.com/periods-with-quotation-marks/

Though this seems to be due to reasons that are no longer relevant: https://style.mla.org/punctuation-and-quotation-marks/

I personally prefer what you suggest but am indifferent. @sedwards-lab do you have opinions here?

sedwards-lab: The American standard rule is that the period be inside the quote. It's illogical from a math/CS perspective, but that's the convention here.

hmontero1205: haha ok nevermind then

j-hui (approved):

PR #49, opened 2021-11-17
Typeclass instantiation

LeoQiao18:
For the initial implementation of typeclass instantiation, we are dealing with typeclass that has a single parameter, no parent typeclass and no default implementations. The approach that we are using is dictionary passing.

An example of transformation looks like the following:
Before:
```sslang
class Collapsible a where
  collapse: a -> Bool
  
instance Collapsible Bool where
  collapse(x: Bool) -> Bool =
    (x: Bool)
  
instance Collapsible Int where
  collapse(x: Int) -> Bool =
    match (x: Int)
      | (0: Int) = (False: Bool)
      | otherwise = (True: Bool)
```

After:
```sslang
type Collapsible a = Collapsible { collapse: a -> Bool }

collapsible_bool: Collapsible Bool =
  Collapsible {
    collapse = func (x: Bool) -> Bool {
      (x: Bool)
    }
  }

collapsible_int: Collapsible Int =
  Collapsible {
    collapse = fun (x: Int) -> Bool {
      match (x: Int)
        | (0: Int) = (False: Bool)
        | otherwise = (True: Bool)
    }
  }
```

This is an example of using the `Collapsible` typeclass above.

Before:
```sslang
collapseToggle[Collapsible a](x: a, led: &Bool) -> () =
  (led: &Bool) <- ((collapse: a -> Bool) (x: a): Bool)

useCollapseToggle(led: &Bool) -> () =
  (led: &Bool) <- ((collapseToggle: [Collapsible a](a, &Bool)) ((1: Int), (led: &Bool)): ())
```

After:
```sslang
collapseToggle(collapsible: Collapsible a, x: a, led: &Bool) -> () =
  (led: &Bool) <- ((collapse: (Collapsible a -> (a -> Bool)) (collapsible: Collapsible a)): (a -> Bool) (x: a): Bool)
```
or, instead of using record accessors, use explicit pattern-matching:
```sslang
collapseToggle: Collapsible a -> a -> &Bool -> () =
  fun (collapsible)
    match collapsible
      | Collapsible { collapse = collapse } =
          fun (x, led)
            led <- collapse x
```


LeoQiao18:
This commit marks the initial completion of lexer, parser, AST, IR, lower-ast, and other adjustments in the codebase for typeclass and instance definitions. The most notable change is the AST change. The `Program` type is now defined as the following:
```haskell
-- | A complete program: a list of declarations
newtype Program = Program [TopDef]
  deriving (Eq, Show)

-- | A top-level definition can be: definition, class or instance
data TopDef = TopDef Definition
            | TopClass TopClass
            | TopInst TopInst
            deriving (Eq, Show)

type TopClass = (TClassId, TVarId, [(VarId, TypAnn)])

type TopInst = (TClassId, TConId, [Definition])

-- | A value definition
data Definition
  = DefFn VarId [Pat] TypFn Expr
  | DefPat Pat Expr
  deriving (Eq, Show)
```

LeoQiao18:
All tests are now passing. Prettyprint is also working.

src/Front/Parser.y
```diff
@@ -77,9 +79,13 @@ indeed semantically) distinct from actual let-bindings (@defLets@).
 They are separated with '||' rather than ';' because the latter somehow causes
 shift-reduce conflicts.
 -}
-defTop                                --> [Definition]
-  : defLet '||' defTop                  { $1 : $3 }
-  | defLet                              { [$1] }
+defTop                                --> [TopDef]
```
j-hui, line 82: This is kind of ugly. Why don't we separate this into:

```
defTops  --> [Definition]
  : defTop '||' defTops { $1 : $3 }
  | defTop { [$1] }
defTop --> Definition
  : defLet { TopDef $1 }
  | defClass { TopClass $1 }
  | defInst { TopInst $1 }
```

Something like that

src/IR/LowerAst.hs
```diff
@@ -48,10 +48,47 @@ lowerType a@(A.TApp _ _      ) = case A.collectTApp a of
...
-lowerProgram (A.Program ds) = return $ I.Program
+lowerProgram (A.Program []) = return $ I.Program
   { I.programEntry = fromString "main"
-  , I.programDefs  = map (first fromJust . lowerDef) ds
+  , I.programDefs  = []
   , I.typeDefs     = []
+  , I.classDefs    = []
+  , I.instDefs     = []
+  }
+lowerProgram (A.Program (A.TopDef d : tds)) =
```
j-hui, line 58: This should be broken up into some kind of map or fold; the amount of code here suggests that this is doing more than it is.

My suggestion is to filter the definition list into three different lists, of defs, classes, and insts, and then map `lower{Def,Class,Inst}` over those. Something like:

```haskell
return $ I.Program
  { I.programEntry = fromString "main"
  , I.programDefs = map lowerTopDef $ filterBy (\case -> { A.TopDef d -> Just d; _ -> Nothing }) ds
  , I.classDefs = map lowerTopDef $ filterBy (\case -> { A.TopClass c -> Just c; _ -> Nothing }) ds
  , I.programDefs = map lowerTopDef $ filterBy (\case -> { A.TopInst i -> Just i; _ -> Nothing }) ds
  -- etc.
  }
  where filterBy f = catMaybes . map f
```

src/Front/Ast.hs
```diff
@@ -22,9 +22,19 @@ type VarId = String
...
+newtype Program = Program [TopDef]
   deriving (Eq, Show)
 
+-- | A top-level definition can be: definition, class or instance
+data TopDef = TopDef Definition
+            | TopClass TopClass
+            | TopInst TopInst
+            deriving (Eq, Show)
+
+type TopClass = (TClassId, TVarId, [(VarId, TypAnn)])
```
j-hui, line 34: These should be records types; it's a clearer what the members represent.

src/IR/IR.hs
```diff
@@ -33,9 +42,23 @@ data Program t = Program
   { programEntry :: VarId
   , programDefs  :: [(VarId, Expr t)]
   , typeDefs     :: [(TConId, TypeDef t)]
+  , classDefs    :: [ClassDef t]
```
j-hui, line 45: let's use a more uniform naming convention. How about `programDefs`, `programClasses`, `programInsts`, and `programTypes`?

src/Front/Ast.hs
```diff
@@ -115,6 +125,18 @@ collectApp t               = (t, [])
 instance Pretty Program where
   pretty (Program defs) = vsep (intersperse emptyDoc $ map pretty defs)
 
+instance Pretty TopDef where
+  pretty (TopDef d) = pretty d
+  pretty (TopClass (tcid, tvid, ms)) =
+    pretty "class" <+> pretty tcid <+> pretty tvid <+> braces
+      (hsep $ punctuate semi $ map
+        (\(vid, typann) -> pretty vid <+> pretty typann)
```
j-hui, line 133: How about `bimap pretty pretty` instead?

src/IR/ClassInstantiation.hs
```diff
@@ -1,15 +1,39 @@
 module IR.ClassInstantiation where
```
j-hui, line 1: You should indicate that this is currently a nop/throw an error if you find any classes or instances.

j-hui (changes requested):
Sorry for taking so long to review this, and thanks for this work! I left behind quite a few nits.

Note that since this changes the IR, this will likely break a lot of other in-flight PRs. I'll re-review and help upstream once some of those are merged in + my current nits are addressed.

src/Front/Ast.hs
LeoQiao18: `bimap pretty pretty` gives `[(Doc, Doc)]`, while we need `[Doc]` here. So I think the lambda is necessary to collapse the tuple.

LeoQiao18: `bimap pretty pretty` gives `[(Doc, Doc)]`, while we need `[Doc]` here. So I think the lambda is necessary to collapse the tuple.

LeoQiao18:
@j-hui Implemented instantiation for class definitions. Currently, I am unable to implement instantiation for instances, because we are lacking ADT expression. The current ADT expression in `src/IR/IR.hs` is the following:
```
data Expr t
  = ...
  | Data DConId t
  | ...
```
Is anyone currently working on this? If not, I will start to think about how to implement this.

j-hui:
> Currently, I am unable to implement instantiation for instances, because we are lacking ADT expression. Is anyone currently working on this? If not, I will start to think about how to implement this.

I'm working on it! Sorry for dragging that out. I'll let you know when that's ready to be merged into your branch. But that's more or less what you need to do. If you want to instantiate a data constructory named `Bar`, the corresponding expression in IR looks like `Data (fromString "Bar") ty` where `ty` is the type of `Bar`. And if `Bar` is not a nullary constructor, i.e., it takes multiple arguments, then the type of `Bar` will be a function that takes those types as arguments. e.g.,:

```
type Foo = Bar | Baz Int // type definition

Bar // nullary data constructor
// becomes Data (fromString "Bar") (TCon (fromString "Foo"))

Baz // unary data constructor
// becomes Data (fromString "Bar") (int 32 `arrow` TCon (fromString "Foo))
```

If you don't want to write `fromString` everywhere, use `{-# LANGUAGE OverloadedStrings #-}`

LeoQiao18:
@j-hui Implemented instantiation for instances. Added a corresponding test. The next step for typeclass instantiation is function that uses typeclasses. This would need to be implemented throughout the whole compiler pipeline: syntax for typeclass usage, IR and typesystem representation, type-checking + type-inference, etc.

j-hui:
RIP typeclasses (for now)

PR #50, opened 2021-11-17
initialize adt c structs

EmilySillars:
I started editing codegen.hs to generate initialization of c structs.
I ran into an issue where I didn't have all the information I needed to determine whether an ADT was an integer or pointer. I think this information needs to be passed into gen expr after the call to genTypeDef (I think  genTypeDef will be called in genTop?), or c macros need to be defined at the same as the adt c struct definitions (genTypeDef should generate C macros). I think the latter would be less invasive but I'd like to hear your thoughts on this.

### DIFF
--- a/src/Codegen/Codegen.hs
+++ b/src/Codegen/Codegen.hs
@@ -436,10 +436,8 @@ genExpr (I.Var n (I.TBuiltin (I.Arrow _ _))) =
 genExpr (I.Var n _) = do
   -- isLocal <- isLocalVar n -- TODO: check for global vs local variable
   return ([cexp|$id:acts->$id:n|], [])
-genExpr (I.Data _ t) = do
-  tmpName <- nextTmp t
-  return ([cexp|$id:tmpName|], [])
-genExpr (I.Lit l t              ) = genLiteral l t
+genExpr (I.Data _ _) = fail "genExpr (I.Data _ _) should never be reached."
+genExpr (I.Lit l t) = genLiteral l t
 genExpr (I.Let [(Just n, d)] b _) = do
   addLocal (n, extract d)
   (defVal, defStms) <- genExpr d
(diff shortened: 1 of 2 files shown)


EmilySillars:
I made changes to 
- Identifiers.hs (add runtime adt/mm related c-code types and identifiers)
- Typdef.hs (generate C-code for each adt and store type info in a lookup table TypeDefInfo)
- Codegen.hs (thread TypeDefInfo through state monad, c-struct initialization and pattern matching)

The C code generated should match draft-v2 branch's c definitions, and uses the following example as reference:
```
// data MyBool = MyTrue int | MyFalse int int
// let a = MyFalse 42 18 

enum MyBool{MyTrue = 0, MyFalse};

void main(){
	ssm_value_t tmp; 
	// ssm_new takes (max size across all data constructors, tag)
	tmp.heap_ptr = (struct ssm_mm*) ssm_new(2, MyFalse);
	((struct ssm_object*)tmp.heap_ptr)->payload[0].packed_val = ssm_int_obj(42);
	((struct ssm_object*)tmp.heap_ptr)->payload[1].packed_val = ssm_int_obj(18);
	ssm_value_t a = tmp;
}
```

sedwards-lab:
This seems like it has more casts than it needs.

```
void main(){
	ssm_value_t tmp; 
	// ssm_new takes (max size across all data constructors, tag)
	tmp.heap_ptr = (struct ssm_mm*) ssm_new(2, MyFalse);
	((struct ssm_object*)tmp.heap_ptr)->payload[0].packed_val = ssm_int_obj(42);
	((struct ssm_object*)tmp.heap_ptr)->payload[1].packed_val = ssm_int_obj(18);
	ssm_value_t a = tmp;
}
```

shouldn't it just be `tmp.heap_ptr->payload[0].packed_val = ssm_int_obj(42);`?
Similarly, shouldn't `ssm_new()` always return an `ssm_value_t` ?

j-hui:
The `draft-v2` branch contains some macros to get rid of pointer casts entirely. You should be able to just write:

```c
void main() {
  struct ssm_object obj = ssm_new(/* max xize */ 2, /* tag */ MyFalse);
  obj->payload[0] = ssm_marshal(42);
  obj->payload[1] = ssm_marshal(18);
  ssm_value_t a = ssm_from_obj(tmp);
}
```

The types of some of these macros are, roughly:

```ocaml
ssm_marshal : int -> ssm_value_t
ssm_from_obj : struct ssm_object * -> ssm_value_t
ssm_to_obj : ssm_value_t -> struct ssm_object *
```

j-hui:
> Similarly, shouldn't ssm_new() always return an ssm_value_t ?

I wrote `ssm_new()` to return a `struct ssm_object *` because the caller still needs to initialize the payloads, so some kind of cast by the caller is just unavoidable. I figured that having `ssm_new()` (and allocating functions) return a more specific type was more self-documenting, so I had it return `struct ssm_object *`. I tried to make the caller perform as many casts as possible because ultimately the caller is the one with the most type information.

EmilySillars:
I've updated codegen.hs / identifiers.hs so (hopefully) the following code will be generated with this example:
```
// data MyBool = MyTrue int | MyFalse int int
// let a = MyFalse 42 18 
enum MyBool{MyTrue = 0, MyFalse};
void main() {
  struct ssm_object* tmp = ssm_new(/* max size */ 2, /* tag */ MyFalse);
  tmp->payload[0] = (ssm_value_t) { .packed_val = ((42) << 1 | 1) };
  tmp->payload[1] = (ssm_value_t) { .packed_val = ((18) << 1 | 1) };
  ssm_value_t a = (ssm_value_t) { .heap_ptr = &(tmp)->mm };
}
```
I tried to have codegen generate the macros itself as much as possible. The code generated should be equivalent to
```
void main() {
  struct ssm_object* tmp = ssm_new(/* max size */ 2, /* tag */ MyFalse);
  tmp->payload[0] = ssm_marshal(42);
  tmp->payload[1] = = ssm_marshal(18);
  ssm_value_t a = ssm_from_obj(tmp);
}
```
Currently working on writing a unit test with the MyBool example in IR. Once I have it in IR I will test my work.

PR #52, opened 2021-11-17
Add minimal annotated version of each regression test

XijiaoLi:


j-hui (approved):
Awesome, thanks @XijiaoLi !

I suppose we still need to annotate the return type for the top-level functons (i.e., `-> ()`) because type inference isn't there yet. Is that right?

XijiaoLi:
> Awesome, thanks @XijiaoLi !
> 
> I suppose we still need to annotate the return type for the top-level functons (i.e., `-> ()`) because type inference isn't there yet. Is that right?

Correct. I actually fixed this using my half type-checker half HM stuff, but since that isn't something that should even exist in the world, I will just leave the type inference as is until we have a fully working HM type inference.

j-hui:
Got it. Feel free to merge!

PR #53, opened 2021-11-17
Upstream ns time base

j-hui:
Upstream lib/ssm to use nanosecond time base


j-hui:
Lemme reopen this with a pre-squashed commit

PR #57, opened 2021-11-18
Fix haddock

j-hui:
For now, this stops haddock from complaining about anything. Some of the documentation is still hastily written, but as we revisit that code it will/should be improved.

I will open this as a full PR once I have taken a look through the actual built site.


j-hui:
there's basically no way this still merges cleanly. I'm going to close this and revisit this issue in a later PR.

PR #58, opened 2021-11-20
Support identifiers uniformly in front-end

j-hui:
This is still very much a work-in-progress, but I'm finally adding support for these two kinds of patterns. In particular, we would like:

- [x] (numeric) literal patterns, e.g., `12`
- [x] Juxtaposed patterns, e.g., `Foo a b`
- [x] Data constructors, e.g., `None` (technically supported, but details are being worked out; see below)

As I'm doing this, I'm also adding support for data constructors. We had already agreed that, following Haskell's convention, data constructors and type constructors should begin with upper case, while data variables and type variables should begin with lower case.

At this time of writing, I have_not_ made this distinction, primarily because it would lead to a lot of code duplication in the front ends of the parser. In fact, distinguishing upper-case and lower-case identifiers is just not _grammatically_ necessary.

My current approach is to give maximum flexibility and parse everything as identifier nodes which just contain a string. So it's technically possible to parse something like this: `let f (a b) = ...`. That's nonsensical, but should be straightforward enough to check later in the compiler. I will write this pass (and tidy up the check and late parse stages of the compiler front-end) before opening up this draft PR.

This makes me think it's not even necessary to distinguish between the two in the IR, i.e., no need to distinguish between `Var` and `Data`. So I might also straighten this out in the IR too, though those codechanges may have implications for others, and may need more careful coordination.


j-hui:
The pass is now complete, but there is still some work to be done to get this actually integrated.

TODOs for myself:

- [ ] Make sure the lowering phase correctly translates to the IR, where Var and Data are differentiated, from AST, where they aren't
- [ ] Add "standard library" identifiers
- [ ] Add tests, making sure to test co-recursion

j-hui:
Opening this PR now since it should be passing tests, but I still need to add tests, plus I need to:

- [ ] figure out a better place to put the standard library identifiers

j-hui:
Started adding some tests, and ended up tainting a bunch of other stuff due to how the tests are being restructured. This is becoming a very undisciplined PR, but these test combinators should help make future tests easier to write.

src/Common/Compiler.hs
```diff
@@ -1,52 +1,68 @@
...
+newtype ErrorMsg = ErrorMsg String
+  deriving Show
+  deriving Semigroup    via String
+  deriving Monoid       via String
+
+instance IsString ErrorMsg where
+  fromString = ErrorMsg
+
+instance Eq ErrorMsg where
+  _ == _ = True
```
j-hui, line 41: By overloading `ErrorMsg`, we can treat all error messages as being equal. Then, when we compare equality on `Error`s, we are just comparing the broader category of error. This allows us to check which kind of an error we are throwing without having to restrict the contents of the error message we report.

Correspondingly, variants of `Error` should be chosen according to the granularity of which classes of errors we consider equal.

src/Common/Identifiers.hs
```diff
@@ -141,3 +150,13 @@ newtype FieldId = FieldId Identifier
 
 -- | A name to be bound; 'Nothing' represents a wildcard, e.g., @let _ = ...@
 type Binder = Maybe VarId
+
```
j-hui, line 153: This is how we classify whether an identifier identifies a constructor or a variable (for both types and data identifiers).

src/Front/Ast.hs
```diff
@@ -1,45 +1,31 @@
...
 
--- | A type class name (e.g., Eq, Ord)
-type TClassId = String
-
--- | A variable name (e.g., x, y, f)
-type VarId = String
-
--- | An operator name (e.g., +, foo)
-type OperatorId = String
+import           Data.List                      ( intersperse )
```
j-hui, line 8: There's no point in creating all these type aliases here when these aren't even being differentiated by the parser. Instead, we use `Identifier` for any such node in the AST.

src/Front/Ast.hs
```diff
@@ -1,45 +1,31 @@
...
   deriving (Eq, Show)
 
 -- | A pattern appearing on the LHS of a definition or match arm
 data Pat
-  = PatWildcard         -- ^ Match anything, i.e., @_@
-  | PatId VarId         -- ^ Bind value to variable, e.g., @v@
-  | PatLit Literal      -- ^ Literal match, e.g., @1@
-  | PatAs VarId Pat     -- ^ Pattern alias, e.g., @a \@ <pat>@
-  | PatTup [Pat]        -- ^ Match on a tuple, e.g., @(<pat>, <pat>)@
-  | PatCon DConId [Pat] -- ^ Match on data constructor, e.g., @Some <pat>@
```
j-hui, line 41: This is not something the parser even recognizes, so there's no point having this node. Instead, this is now expressed as a `PatApp` of `PatId`.

src/Front/Check.hs
```diff
@@ -1,46 +0,0 @@
-module Front.Check where
```
j-hui, line 1: Deleted this since it's something that's not really consistent with the presence of type inference.

We can bring this back once we add warning infrastructure.

src/Front/Identifiers.hs
```diff
@@ -0,0 +1,25 @@
+{-# LANGUAGE OverloadedStrings #-}
```
j-hui, line 1: This module is where we will put information on reserved/builtin identifiers. (oops I forgot documentation, will add)

src/Front/ParseOperators.hs
```diff
@@ -1,22 +1,24 @@
...
 
-import qualified Data.Map.Strict               as Map
 import           Front.Ast                      ( Definition(..)
                                                 , Expr(..)
+                                                , Fixity(..)
                                                 , OpRegion(..)
                                                 , Program(..)
                                                 )
 
-data Fixity = Infixl Int String
```
j-hui, line 15: This belongs in `Ast.hs` since eventually we will want to add fixity declarations etc. in the language

src/Front/ParseOperators.hs
```diff
@@ -85,4 +87,6 @@ parseExprOps fixity = rw
   rewrite f (Assign     p  e ) = Assign p (f e)
   rewrite f (Constraint e  t ) = Constraint (f e) t
   rewrite f (Seq        e1 e2) = Seq (f e1) (f e2)
+  rewrite f (Lambda     ps b ) = Lambda ps (f b)
```
j-hui, line 90: These are bugs... smh we weren't recursing on all expressions

src/Front/Parser.y
```diff
@@ -127,8 +138,8 @@ patAtom                               --> Pat
   : '_'                                 { PatWildcard }
   | '(' patTups ')'                     { normalize id PatTup $2 }
   | '(' ')'                             { PatLit LitEvent }
+  | int                                 { PatLit $ LitInt $1 }
```
j-hui, line 141: We now support integer patterns, as well as data constructor patterns via `id`.

Something like `f ((X y) z) = ...` is syntactically valid but we will catch it later in `Scope.hs`

src/Front/Parser.y
```diff
@@ -243,7 +254,7 @@ matchArms                             --> [(Pat, Expr)]
 
 -- | Each individual arm of a pattern match.
 matchArm                              --> (Pat, Expr)
-  : pat '=>' '{' expr '}'               { ($1, $4) }
+  : pat '=' '{' expr '}'                { ($1, $4) }
```
j-hui, line 257: Reverting pattern-match and lambda syntax to use `=` as connective instead of `=>`, for consistency.

src/Front/Parser.y
```diff
@@ -297,9 +304,12 @@ normalize _ f xs = f xs
...
-categorizeDef :: [Pat] -> TypFn -> Expr -> Definition
-categorizeDef [pat] (TypProper typ) = DefPat $ PatAnn typ pat
-categorizeDef [pat] TypNone = DefPat pat
-categorizeDef (PatId f:args) fAnn = DefFn f args fAnn
-categorizeDef _ _ = error "Syntax error in definition"
+categorizeDef :: [Pat] -> TypFn -> Expr -> Alex Definition
+categorizeDef [pat]          (TypProper typ) = return . DefPat (PatAnn typ pat)
+categorizeDef [pat]          TypNone         = return . DefPat pat
+categorizeDef (PatId f:args) fAnn            = return . DefFn f args fAnn
+categorizeDef (_:_)          _               =
```
j-hui, line 311: Instead of throwing `error`, this now correctly throws a `Compiler.Pass` error.

src/Front/Scanner.x
```diff
@@ -114,20 +118,25 @@ tokens :-
...
 -- | User-facing syntax error.
 syntaxErr :: String -> Alex a
-syntaxErr s = alexError $ "Syntax error: " ++ s
+syntaxErr s = alexError $ "_s:" ++ s
 
--- | Internal compiler error for unreachable code.
-internalErr :: String -> Alex a
-internalErr s = alexError $ "Internal error: " ++ s
+liftErr :: String -> Error
+liftErr ('_':'i':':':e) = UnexpectedError $ fromString e
```
j-hui, line 137: Within the alex monad, errors are thrown as strings, so we need to parse them out into `Compiler.Error`

src/Front/Scanner.x
```diff
@@ -419,10 +428,10 @@ collectStream = do
...
 
 -- | Extract a token stream from an input string.
-scanTokens :: String -> Either ErrorMsg [Token]
-scanTokens s = runAlex s collectStream
+scanTokens :: String -> Pass [Token]
+scanTokens = liftEither . first liftErr . flip runAlex collectStream
 
 -- | Extract a stream of token types (without span) from an input string.
-scanTokenTypes :: String -> Either ErrorMsg [TokenType]
+scanTokenTypes :: String -> Pass [TokenType]
```
j-hui, line 435: These are now first-class `Compiler.Pass`es

src/Front/Scope.hs
```diff
@@ -0,0 +1,309 @@
...
+
+Data variables are @switch@, @b@, @x@, and @y@; data constructors are @True@,
+@False@, @Left@, and @Right@. Type variables are @t@, @u@, @a@, and @b@; type
+constructors are @Bool@ and @Either@.
+
+The grammar as it appears in the parser does not actually distinguish between
+any of these kinds of identifiers, so it is the responsibility of this module to
+check that, for example, a data constructor does not appear where a data
+variable is expected, e.g., @let F x = e@, or vice versa, e.g., @let f X = e@.
+-}
```
j-hui, line 47: This documents a relevant (perhaps the main) contribution of this PR

src/IR/LowerAst.hs
```diff
@@ -102,7 +108,8 @@ propogate it elsewhere. This is possible because the IR's type annotations
 the join operation.
 -}
 lowerExpr :: A.Expr -> AnnotationK -> I.Expr I.Type
-lowerExpr (  A.Id  v    ) k = I.Var (fromString v) (k I.untyped)
+lowerExpr (A.Id v) k | isCons v  = I.Data (fromId v) (k I.untyped)
```
j-hui, line 111: Here we differentiate between data constructors and variables, and lower to different IR nodes.

src/IR/TypeInference.hs
```diff
@@ -240,4 +241,4 @@ ann2Class (Ann.TBuiltin bty) = case bty of
 ann2Class (Ann.TCon tid tys) = do
   t <- mapM anns2Class tys
   return $ Classes.TCon tid t
-ann2Class (Ann.TVar tidx   ) = return $ Classes.TVar tidx
+ann2Class (Ann.TVar _tid   ) = throwError $ Compiler.UnexpectedError $ fromString "Do not know how to translate type variable yet"
```
j-hui, line 244: We will fix this in #60 when adding support for (quantified) type variable syntax

test/lib/Sslang/Test.hs
```diff
@@ -0,0 +1,85 @@
+{- | Defines some common utilities for writing Hspec tests.
```
j-hui, line 1: This test library makes tests much nicer to write and read, and removes the boilerplate due to imports.

test/semant/Tests/ScopingSpec.hs
```diff
@@ -0,0 +1,263 @@
+{-# LANGUAGE QuasiQuotes #-}
```
j-hui, line 1: This test suite tests the scoping, and includes plenty of example programs

It also demonstrates that now we are able to test that certain kinds of programs fail in a specific way, so we don't get false positives (e.g., we are trying to test that the compiler catches type errors, but it actually fails due to a scope error).

test/semant/Tests/ScopingSpec.hs
```diff
@@ -0,0 +1,263 @@
...
+        foo x =
+          let foo' y = x
+          foo'
+        bar a b =
+          foo b a
+      |]
+    it "rejects undefined references" $ do
+      doScope [here|
+        f = x
+      |] `shouldFailWith` ScopeError "x is undefined"
```
j-hui, line 36: `"x is undefined"` here is an error message. Since error messages are always equal, we can write some descriptive error message about how we expect the test to fail, without worrying about the test being thwarted due to a failed string comparison.

j-hui (commented):
Highlights of this PR:
- More uniform error-reporting across entire compiler
- Parser and lexer now throw compiler errors that we can handle and report from
- Pared down number of test suites
- Uniformly use `Identifier` across AST instead of naked strings
- Fix some bugs in ParseOperators
- Perform scoping checks
- Better unit tests that also check for

j-hui:
Just added support for warnings --- now we can just `warn $ KindOfWarning "warn message"` from anywhere within the `Pass` monad to throw a warning:

```
$ stack exec sslc -- - >/dev/null <<-EOF
f (x: Int) -> Int =
  let (x: Int) = 3
  x
EOF
[NameWarning (ErrorMsg "shadowing variable: 'x'")]
```
Right now it just dumps the default `Show` instance, which is ugly, but now we can easily fine-tune the CLI reporting for warnings and errors by creating a `Pretty` instance and using that

src/Front/Scope.hs
```diff
@@ -0,0 +1,311 @@
...
+dataDecl i = do
+  info <- asks $ M.lookup i . dataMap
+
+  -- A data constructor cannot be defined inside of a pattern.
+  when (isCons i && not (inScope info)) $ do
+    throwError $ ScopeError $ "Data constructor is out of scope: " <> showId i
+
+  -- A data variable can usually be shadowed, but we want warn about it.
+  when (isVar i && inScope info) $ if canShadow info
+    then warn $ NameWarning $ "shadowing variable: " <> showId i
```
j-hui, line 171: Just added support for warnings --- now we can just `warn $ KindOfWarning "warn message"` from anywhere within the Pass monad to throw a warning:

```
$ stack exec sslc -- - >/dev/null <<-EOF
f (x: Int) -> Int =
  let (x: Int) = 3
  x
EOF
[NameWarning (ErrorMsg "shadowing variable: 'x'")]
```

Right now it just dumps the default Show instance, which is ugly, but now we can easily fine-tune the CLI reporting for warnings and errors by creating a Pretty instance and using that

src/Front/Parser.y
hmontero1205: Does sslang support booleans all the way through? Is it worth throwing in bool literals while you're already here?

src/Front/Parser.y
hmontero1205: ..consistency with what? this isn't assignment or a binding..

src/Front/Parser.y
```diff
@@ -270,16 +281,12 @@ exprPar                               --> [Expr]
...
 {
 -- | What to do upon encountering parse error.
---
--- TODO does this really just throw an uncaught exception on parse error, or
--- does Happy catch that for us?
-parseError :: Token -> a
-parseError (Token (sp, _)) =
-    error $ show (pretty sp) ++ ":Syntax error"
+parseError :: Token -> Alex a
+parseError (Token (sp, _)) = syntaxErr $ "near " ++ show (pretty sp)
```
hmontero1205, line 285: near? not "in" or something? i am not fully sure what span looks like, but i'd assume the error is somewhere inside the span? unless you are saying near just to be safe since the code looks very different the further down the compiler pipeline you go?

src/Front/Parser.y
```diff
@@ -270,16 +281,12 @@ exprPar                               --> [Expr]
...
 
 -- | Parse a 'String' and yield a 'Program' or an 'ErrorMsg' if unsuccessful.
```
hmontero1205, line 287: is this docstring still accurate? that was obvious to me when it was `Either ErrorMsg Program`.

src/Front/Scanner.x
hmontero1205: cursed but understandable :(

src/Front/Scope.hs
```diff
@@ -0,0 +1,311 @@
...
+    else
+      throwError $ NameError $ "Cannot bind identifier shadowing : " <> showId i
```
hmontero1205, line 173: stray space after shadowing?

test/lib/Sslang/Test.hs
```diff
@@ -0,0 +1,85 @@
...
+
+-- | Expect that some 'Pass' must produce a particular value.
+shouldProduce :: (HasCallStack, Show a, Eq a) => Pass a -> a -> Expectation
+shouldProduce actual expected = case runPass actual of
+  Right (a, _) -> a `shouldBe` expected
+  Left  a -> assertDifference "Expected success but encountered error"
+                              (show a)
+                              (show expected)
+
+-- | Expect that some 'Pass' must produce some expected value.
```
hmontero1205, line 64: this docstring is misleading imo. it implies the RHS has already been evaluated, which isnt the case.

maybe something like..
Expect that some 'Pass' must produce the same value as another 'Pass'

test/lib/Sslang/Test.hs
```diff
@@ -0,0 +1,85 @@
...
+    assertFailure $ "Encountered error when evaluating expectation: " ++ show e
+  (Left a, Right e) ->
+    assertDifference "Expected success but encountered error" (show a) (show e)
+  (Right (a, _), Right (e, _)) -> a `shouldBe` e
+
+-- | Expect that some 'Pass' should fail with the given 'Error'.
+shouldFail :: (HasCallStack, Show a) => Pass a -> Expectation
+shouldFail actual = case runPass actual of
+  Right (a, _) ->
+    assertDifference "Did not encounter expected error" (show a) "Some error"
```
hmontero1205, line 77: assertFailure seems more appropriate.. you dont have a real value for the RHS.

hmontero1205 (commented):
John, I think this PR got way too big. This did a lot more than just add support for data constructors and literal patterns XD

Your comments did guide me through your changes, which seem good to me. I left some comments here and there, but no real request for changes. LMK if you want me to double check any particular area.

src/Front/Parser.y
j-hui: the syntax werror isn't inside of span, since span just describes the extent of a single token. And it only describes the unexpected token:

```
$ stack exec sslc -- -
x =
  let y = blah blah
ParseError (ErrorMsg "near 3:1     [24+0]  ")
```

The right error message here is that there is an unfinished expression introduced by the `let` keyword, but the parse error complains about the newline at the end of the file.

I'll change this to "at" but we should have very little confidence in this generated parser at providing useful error messages.

src/Front/Scope.hs
j-hui: good catch

test/lib/Sslang/Test.hs
j-hui: good point

j-hui:
Thanks for the review Hans! And yeah this PR definitely got out of hand; I wanted to have this properly tested but was too impatient to wait for other infrastructural PRs to be merged in. I addressed all of your concerns.

hmontero1205 (approved):

PR #59, opened 2021-11-26
Fix inconsistent type annotation in doubleblink

XijiaoLi:
Since the type of `par` should be a tuple, the annotated return type of `main` function (`()`) is inconsistent with the actual type of its function body (`(), ()`). Since currently we not yet support handling tuple type inside codegen, add a dummy `()` at the end of function body so that its type matches the annotated return type.

### DIFF
--- a/regression-tests/tests/doubleblink.ssl
+++ b/regression-tests/tests/doubleblink.ssl
@@ -18,3 +18,4 @@ fast(led : &Int) -> () =
 main(led : &Int) -> () =
   par ((slow: &Int -> ()) (led: &Int): ())
       ((fast: &Int -> ()) (led: &Int): ())
+  ()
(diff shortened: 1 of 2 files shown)


j-hui (approved):
Thanks for the timely fix @XijiaoLi !

Just to double check, why was it passing the test before? Seems like it could be a bug in type-checking.

XijiaoLi:
> Thanks for the timely fix @XijiaoLi !
> 
> Just to double check, why was it passing the test before? Seems like it could be a bug in type-checking.

Since I just naively threw away the type annotation for `par` and replace it with tuple at [here](https://github.com/ssm-lang/sslang/blob/b02e7257fe0cd95131de686ccd546b29b06e4157/src/IR/TypeInference.hs#L190). Indeed a bug in the type checker though lol.

j-hui:
@XijiaoLi got it. Let's make sure to patch that up on the next iteration. By that point I hope to have codegen for tuples working (:

PR #60, opened 2021-12-02
Implement Hindley Milner type inference version 1

XijiaoLi:
In this PR:
- implement the very first version of HM type inference
  - [x] add type scheme.
  - [x] implement all the basic unification functionalities.
  - [x] verify the correctness of the implementation with tests.

Next steps:
  - instead of blindly throwing away any type annotations, carry them into the type inference (first collapse and then unify with other inferred type).
  - rename the quantified type scheme.
  - integrate with type class stuff.


src/Common/Identifiers.hs
```diff
@@ -86,17 +88,18 @@ instance Show TVarId where
 
 -- | de Bruijn index for type variables, e.g., @'0@
 newtype TVarIdx = TVarIdx Int
-  deriving Eq
+  deriving (Eq, Ord)
 
 instance Show TVarIdx where
-  show (TVarIdx i) = "'" ++ show i
+  show (TVarIdx i) = "t" ++ show i
```
j-hui, line 94: Just curious, was there a reason for this change?

test/typeinference/Tests/InferProgramSpec.hs
```diff
@@ -77,3 +87,112 @@ spec = do
...
+  For now, I hardcode all the expected results. We should fix that later.
+  Uncomment the print statement to output the inference result for each function.
+  -}
+  it "infers id function" $ do
+    let minimal = parseProgram $ unlines
+          [ "id(x) ="
+          , "  x"
+          ]
+        typedMinimal = lowerAndInfer minimal
+        typedPString = show typedMinimal
```
j-hui, line 102: We shouldn't be comparing the string representation of the inferred type when we can just compare the type representation directly.

j-hui: oh wait yikes I see why you're doing this. This seems kind of fragile.

Why not compare an unannotated and fully annotated version? e.g., compare `addOne` to:

```
addOne x : Int -> Int = ((x: Int) + 1):Int
```

I realize that there isn't yet syntax for type variables, so I'll get working on that.

test/typeinference/Tests/InferProgramSpec.hs
```diff
@@ -77,3 +87,112 @@ spec = do
...
+  it "infers tricky program #3:`f` should has type `a -> (b -> b)` not `a -> (a -> a)`" $ do
+    let minimal = parseProgram $ unlines
+          [ "f x ="
+          , "  let g x = x"
+          , "  g g"
+          ]
+        typedMinimal = lowerAndInfer minimal
+        typedPString = show typedMinimal
+    -- print typedMinimal
+    typedPString `shouldContain` "Var g (TBuiltin (Arrow (TVar t4) (TVar t4)))"
```
j-hui, line 161: Fishing for the index of `TVar`s seems really fragile. I think we should write a custom comparison method for IR trees that considers identifiers modulo alpha renaming.

j-hui (commented):
This is awesome, thanks Xijiao! I haven't yet taken a close look through `HM.hs`; let's talk about that at our meeting today.

My comments are primarily about the tests. Some of my concerns there are contingent on my syntax additions in #58, which I will work on today. Sorry for the delay.

j-hui:
@XijiaoLi I helped merge in the latest changes from `main`.

Note that your changes broke one of Hans's lambda lifting tests. His tests were overconstrained, i.e., the type of the expression was more general than what was specified in the type annotation. Once you adjust type inference to account for that, we can remove the `pendingWith`

XijiaoLi:
Actually, the ["lifts nested lambdas with free variables"](https://github.com/ssm-lang/sslang/blob/b0c79145010e079e98268ed4c821b6901343615a/test/trans/Tests/LiftProgramLambdasSpec.hs#L66) test failed because of a bug in the `findRoot` function: it did not correctly handle the case when the value of a `TVar` inside the union find tree is a built-in type.

For example, after unification, we want to replace the temporary `TVar t8`. We find `TVar t8 : Arrow (TVar t4) (TVar t7)` in the uft. The `findRoot` would simply use `Arrow (TVar t4) (TVar t7)` to replace `t8`, instead of further resolving `TVar t4` `TVar t7` using their corresponding root values in the uft. So there are some `TVar` left in the final type, even through they can be replaced with some concrete types.

src/IR/HM.hs
```diff
@@ -0,0 +1,490 @@
+{-# LANGUAGE DerivingVia #-}
```
j-hui, line 1: Can you include module documentation outlining the overall architecture of what's going on here/how to navigate this module?

j-hui (approved):
Looks great, this is awesome @XijiaoLi !

Before you merge, I left a few nits here and there, mostly to clarify a few things about organization. Also, could you make it clear somewhere what's left, i.e., what are the things you don't yet support? If possible, the best thing to do would be to add them to your test suite as `pendingWith` tests (which won't be run).

j-hui:
Oh also, does this still use the old typechecker from before?

XijiaoLi:
> Oh also, does this still use the old typechecker from before?

@j-hui It doesn't. Should I make it run through both (i.e. throws an error if either one fails)?

test/lib/Sslang/Test.hs
```diff
@@ -101,6 +101,15 @@ shouldNotPassAs actual unexpected = do
   when (e == a) $ assertFailure $ unlines
     ["Expected inequality but found equivalent:", show a, "----", show e]
 
+-- | Expect that some 'Pass' successfully produces a different value from another.
+shouldPassButNotAs :: (HasCallStack, Show a, Eq a) => Pass a -> Pass a -> Expectation
```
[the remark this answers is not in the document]
XijiaoLi: Ah yes, `shouldPassButNotAs` has the same effect as your `shouldNotPassAs`. I forgot to delete this function after merging in the main branch. Let me change the function name to `shouldPassButNotAs` then!

j-hui:
Throw an error when _both_ report failure. My understanding is that both currently throw false negatives (complain about type-safe programs), right? Let's make sure that even if HM doesn't work for some programs, we can always fallback to what we get out of type annotations.

XijiaoLi:
> Throw an error when _both_ report failure. My understanding is that both currently throw false negatives (complain about type-safe programs), right? Let's make sure that even if HM doesn't work for some programs, we can always fallback to what we get out of type annotations.

@j-hui I honestly don't know what is a good way to combine the two checks xD. Currently I am calling the function from the old type checker module inside `HM.hs` and use `runPass` to see if it fails. Could you please take a look at what I did [here](https://github.com/ssm-lang/sslang/blob/xijiao-hm/src/IR/HM.hs#L96) and see if it's reasonable before I merge?

j-hui:
For combining the passes, my recommendation is to do it in the IR top-level module, here: https://github.com/ssm-lang/sslang/blob/105e96165a7e072d11b736a3af1d123ce3ee91d2/src/IR.hs#L78

You could also add command line options to only do one or the other.

XijiaoLi:
@j-hui Can you please double check my latest commits before I merge? I have moved the checking to the IR top-level module and also added command line options for choosing which algorithm to run.

j-hui (approved):
Yepp that's exactly what I had in mind, thanks Xijiao! LGTM

PR #63, opened 2021-12-16
Reorganize compiler front-end

j-hui:
The previous organization made it rather annoying to add new compiler passes, since the entire compiler pipeline was strung together in `app/Main.hs`. Instead, this pushes the minutiae of the compiler stages into `src/{Front,IR,Codegen}.hs`, and leaves `app/Main.hs` to handle CLI parsing and file I/O. In the future, we no longer need to modify `app/Main.hs` to add compiler passes. This design also allows passes to be easily reused in unit tests.

This PR also comes with improvements to the organization and presentation of the unit tests, using a heredoc quasiquoter.


j-hui:
Requesting review from Hans and Xijiao since this affects test cases you have written

XijiaoLi (approved):
LGTM. Thanks!

src/IR.hs
```diff
@@ -1,43 +1,96 @@
...
+options =
+  [ Option ""
+           ["dump-ir"]
+           (NoArg $ setMode DumpIR)
+           "Print the IR immediately after lowering"
+  , Option ""
+           ["dump-ir-typed"]
+           (NoArg $ setMode DumpIRTyped)
+           "Print the fully-typed IR after type inference"
+  , Option ""
```
hmontero1205, line 57: This is meaning an option for `DumpIRLifted`?

src/IR.hs
```diff
@@ -1,43 +1,96 @@
...
+system.
 
-inferDrops :: I.Program a -> Compiler.Pass (I.Program a)
-inferDrops = return
+TODO: This is probably not necessary in the future.
+-}
+poly2Flat :: Options -> I.Program Poly.Type -> Pass (I.Program Flat.Type)
+poly2Flat opt ir = do
+  irLifted <- liftProgramLambdas ir
+  when (mode opt == DumpIRLifted) $ dump irLifted
```
hmontero1205, line 90: Why is lifting not its own operation here?

hmontero1205 (commented):
I left some questions -- not sure if what you did with lifting stuff was intentional.

src/IR.hs
hmontero1205: Like why is it no longer a top-level operation?

j-hui: Why does it need to be? I'm encapsulating it here so that there doesn't need to be all this extra boilerplate invoking all the subphases.

src/IR.hs
j-hui: Oops yeah, I'll add that in.

src/IR.hs
hmontero1205: I guess it doesn't have to be, I suppose I just saw it that why because I implemented it. I see now why your organization makes sense, just wanted to ask why this change was made.

j-hui: Gotcha. Yeah I did that because I saw that there was a lot of boilerplate all over the place, stitching together the same compiler pipeline in different places.

hmontero1205 (approved):

PR #65, opened 2021-12-29
Add type definition syntax

j-hui:
Needed for #46, #49, #50, and #60---this syntax will allow user-defined ADTs, so that we can actually test those language features with real programs.

Will conflict a little with #49, which also makes changes to AST, parser, and related components to add syntax for declaring typeclases and defining their instances. @LeoQiao18 , I'm happy to help resolve those merge conflicts.

This PR is still a WIP, but should be done within the next day. Currently, I still have to:

- [x] Extend the scoping pass to add and check user-defined type constructors, and add user-defined data constructors to the data namespace.
- [x] Add parser tests to make sure that this syntax is supported.

### DIFF
--- a/src/Front/Ast.hs
+++ b/src/Front/Ast.hs
@@ -14,14 +14,14 @@ newtype Program = Program [TopDef]
 -- | A top-level definition.
 data TopDef
   = TopDef Definition     -- ^ Bind a (data) value to a variable
-  | TopType TypeDef       -- ^ Declare a type
+  | TopType TypeDef       -- ^ Define an algebraic data type
   deriving (Eq, Show)
 
--- | A type definition.
+-- | An algebraic data type definition.
 data TypeDef = TypeDef
   { typeName     :: Identifier      -- ^ The name of the type, e.g., @Option@
   , typeParams   :: [Identifier]    -- ^ List of type parameters, e.g., @a@
-  , typeVariants :: [TypeVariant]   -- ^ List of variants/data constructors
+  , typeVariants :: [TypeVariant]   -- ^ List of variants, e.g., @Some@, @None@
   }
   deriving (Eq, Show)
 
--- a/src/Front/Ast.hs
+++ b/src/Front/Ast.hs
@@ -7,26 +7,29 @@ import           Common.Pretty
 
 import           Data.List                      ( intersperse )
 
--- | A complete program: a list of declarations
+-- | A complete program: a list of top-level definitions.
 newtype Program = Program [TopDef]
   deriving (Eq, Show)
 
+-- | A top-level definition.
 data TopDef
-  = TopDef Definition
-  | TopType TypeDef
+  = TopDef Definition     -- ^ Bind a (data) value to a variable
+  | TopType TypeDef       -- ^ Declare a type
   deriving (Eq, Show)
 
+-- | A type definition.
 data TypeDef = TypeDef
-  { typeName     :: Identifier
-  , typeParams   :: [Identifier]
-  , typeVariants :: [TypeVariant]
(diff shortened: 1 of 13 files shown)


src/IR/LowerAst.hs
```diff
@@ -54,11 +54,31 @@ lowerType a@(A.TApp _ _) = case A.collectTApp a of
 
 -- | Lower an AST 'Program' into IR.
 lowerProgram :: A.Program -> Compiler.Pass (I.Program I.Type)
-lowerProgram (A.Program ds) = return $ I.Program
-  { I.programEntry = fromString "main"
-  , I.programDefs  = map (first fromJust . lowerDef) ds
-  , I.typeDefs     = []
-  }
+lowerProgram (A.Program ds) = do
```
j-hui, line 57: Here I lower the type definitions to IR

src/Front/Scope.hs
```diff
@@ -186,13 +218,53 @@ dataRef i = do
...
 -- | Check the scoping of a 'A.Program'.
 scopeProgram :: A.Program -> Pass ()
-scopeProgram (A.Program ds) = runScopeFn $ scopeDefs ds $ return ()
+scopeProgram (A.Program ds) = runScopeFn $ do
+  scopeTypeDefs tds $ scopeDefs dds $ return ()
+ where
+  tds = mapMaybe A.getTopTypeDef ds
+  dds = mapMaybe A.getTopDataDef ds
+
+-- | Check the scoping of a set of type definitions.
```
j-hui, line 235: Here I check scoping for identifiers inside of the type definitions

j-hui (commented):
In this "review" I outline the main contributions of this PR. I had to touch more files than necessary because my PR modified the AST; in particular top-level entities now also encompass type definitions.

src/Front/Ast.hs
```diff
@@ -7,11 +7,29 @@ import           Common.Pretty
...
+data TopDef
+  = TopDef Definition     -- ^ Bind a (data) value to a variable
+  | TopType TypeDef       -- ^ Declare a type
+  deriving (Eq, Show)
+
+-- | A type definition.
+data TypeDef = TypeDef
+  { typeName     :: Identifier      -- ^ The name of the type, e.g., @Option@
+  , typeParams   :: [Identifier]    -- ^ List of type parameters, e.g., @a@
+  , typeVariants :: [TypeVariant]   -- ^ List of variants/data constructors
```
XijiaoLi, line 24: Can we add some examples for `typeVariants` too? Maybe `None` and `Some`, in the context of `Option`?

XijiaoLi (commented):
All the changes related to types look good to me. Thanks John!

src/Front/Ast.hs
```diff
@@ -8,7 +8,22 @@ import           Common.Pretty
 import           Data.List                      ( intersperse )
 
 -- | A complete program: a list of declarations
-newtype Program = Program [Definition]
+newtype Program = Program [TopDef]
```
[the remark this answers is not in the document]
j-hui: Good call

src/Front/Ast.hs
j-hui: will do

src/Front/Ast.hs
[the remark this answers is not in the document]
j-hui: Actually that info is slightly redundant---it should be pretty clear from the definition of `TopDef` what the different kinds of top-level entitites may exist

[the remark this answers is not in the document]
j-hui: I expanded the comments there a little bit instead

XijiaoLi (approved):
Thanks John!

PR #69, opened 2021-12-31
Cache Haskell setup

### LINKED ISSUE #64: CI takes a really long time to run
For example, in this run, it takes 2 whole minutes to set up Stack: https://github.com/ssm-lang/sslang/runs/4577980018?check_suite_focus=true . I'm not sure why it needs to take this long (it takes about a quarter of the time to build our compiler, rebuild our compiler, and run all tests). Most of the time is apparently spent processing 155600 (!) cabal files.

Not very important at the moment, but it would be nice to figure out why it's taking so long.

j-hui:
Fix #64

(or at least an attempt to do so)


j-hui:
Good lord, Stack produces 1.2GB of... stuff. This cache certainly shaves down the setup time, but importing the cache itself somehow takes 1min. But it's still some kind of an improvement.

j-hui:
I added in the stack lock file to see if that changes things

.github/workflows/test.yml
```diff
@@ -1,7 +1,9 @@
-name: Tester
+name: Test
 
 on:
   push:
+    branches:
+      - main
```
XijiaoLi, line 6: Don't we want to run the test on the commits that we push to our own branch?

j-hui: This runs on _pushes_ to `main`, but will still run on all pull requests. So as long as you open a PR (either a full PR or a draft), it will work like it used to. For instance, see this PR; there are multiple check marks throughout the PR, for each separate new commit I pushed to this PR.

This has two benefits: (1) tests won't be run on random orphan branches we create that we never start PRs from; (2) it won't run the same CI action twice when we open a PR, as it currently does (according to the configuration on `main`).

XijiaoLi (approved):
Awesome, thanks John!

PR #70, opened 2022-01-03
Use scope trace to name lifted lambdas

hmontero1205:


j-hui (changes requested):
This is a great start, but I have a few changes to request:

- Use `Identifier` instead of raw `String`s. If you need to concatenate `Identifier`, you can use `<>` (it is an instance of the `Semigroup` and `Monoid` typeclasses). You can use `{-# LANGUAGE OverloadedStrings -#}` to make it such that any string literal gets automatically converted using `fromString`.
- I'm not sure `trail` needs to be a list of identifiers; we can just maintain it as a single identifer that is prefixed to any fresh name. The only instance where that context might be important is for reporting diagnostics, where a line + column number would be more accurate anyway.
- Instead of taking an `Identifier` argument in `descend`, take an `m :: Identifier -> Identifier` function that modifies the existing `trail`. To append to `trail`, pass `(<> suffix)` as `m`, and to replace `trail` entirely, pass `(const $ lamName <> "_")` (in particular, you should do this when descending into a lambda).

hmontero1205:
@j-hui I intentionally used a list instead of a single string for the trail because I think using strings (and Identifier) overcomplicates it. I like being able to use `intercalate "_"` in exactly one place instead of having to account for the delimiter everywhere.

hmontero1205:
@j-hui modifying the code to use Identifier instead of String was trickier than I thought it would be.. please see d9b58d5

src/IR/LambdaLift.hs
```diff
@@ -85,17 +93,26 @@ addCurrentScope :: I.VarId -> LiftFn ()
 addCurrentScope v =
   modify $ \st -> st { currentScope = S.insert v $ currentScope st }
 
--- | Get a fresh variable name
-getFresh :: LiftFn String
+-- | Create a string composed of the scope trail and a variable name.
+prependTrail :: Identifier -> LiftFn Identifier
+prependTrail cur = do
+  curTrail <- gets trail
+  return $ foldr (<>)
```
j-hui, line 100: Replace `foldr .. ..` with `mconcat` (from `Data.Monoid`)

j-hui: You can also replace `Identifier "_"` with just `"_"`; `OverloadedStrings` inserts a `fromString` for you

j-hui: This entire few lines can be reduced to  `return $ mconcat $ intersperse "_" $ reverse $ cur : curTrail`

src/IR/LambdaLift.hs
```diff
@@ -85,17 +93,26 @@ addCurrentScope :: I.VarId -> LiftFn ()
...
+                 (Identifier "")
+                 (intersperse (Identifier "_") (reverse (cur : curTrail)))
+
+-- | Construct a fresh variable name for a new lifted lambda.
+getFresh :: LiftFn Identifier
 getFresh = do
   curCount <- gets anonCount
+  let anonName = Identifier "anon" <> fromString (show curCount)
```
j-hui, line 108: Remove `Identifier`

src/IR/LambdaLift.hs
```diff
@@ -85,17 +93,26 @@ addCurrentScope :: I.VarId -> LiftFn ()
...
   modify $ \st -> st { anonCount = anonCount st + 1 }
-  return $ "anon" ++ show curCount
+  return anonName
 
 -- | Store a new lifted lambda to later add to the program's top level definitions.
-addLifted :: String -> I.Expr Poly.Type -> LiftFn ()
+addLifted :: Identifier -> I.Expr Poly.Type -> LiftFn ()
 addLifted name lam =
-  modify $ \st -> st { lifted = (I.VarId (Identifier name), lam) : lifted st }
+  modify $ \st -> st { lifted = (VarId name, lam) : lifted st }
```
j-hui, line 115: Replace `VarId` with `fromId`

src/IR/LambdaLift.hs
```diff
@@ -150,11 +177,12 @@ liftProgramLambdas p = runLiftFn $ do
 -- | Given a top-level definition, lift out any lambda definitions.
 liftLambdasTop
   :: (I.VarId, I.Expr Poly.Type) -> LiftFn [(I.VarId, I.Expr Poly.Type)]
-liftLambdasTop (v, lam@(I.Lambda _ _ t)) = do
+liftLambdasTop (v@(VarId lamIdent), lam@(I.Lambda _ _ t)) = do
```
j-hui, line 180: You don't need to match on `VarId` here; instead, just use `fromId` on `v`

j-hui: You also don't need to bind `t` here; you can just use `extract lam`

src/IR/LambdaLift.hs
```diff
@@ -182,24 +210,26 @@ liftLambdas (I.Prim p exprs t) = do
...
+    liftLambdas body
   let liftedLam = I.makeLambdaChain
         (map (B.first Just) (M.toList lamFreeTypes) ++ vs')
         liftedLamBody
-  freshName <- getFresh
-  addLifted freshName liftedLam
+  addLifted fullName liftedLam
   return $ foldl applyFree
-                 (I.Var (I.VarId (Identifier freshName)) (extract liftedLam))
+                 (I.Var (VarId fullName) (extract liftedLam))
```
j-hui, line 223: replace `VarId` with `fromId`

src/IR/LambdaLift.hs
```diff
@@ -209,20 +239,31 @@ liftLambdas (I.Match s arms t) = do
 liftLambdas lit@I.Lit{}  = return lit
 liftLambdas dat@I.Data{} = return dat
 
+-- | Entry point for traversing let bindings.
+liftLambdasInLet :: (I.Binder, I.Expr Poly.Type) -> LiftFn (I.Expr Poly.Type)
+liftLambdasInLet (b, expr) = do
+  let bindingName = maybe "wildcard" ident b
```
j-hui, line 245: Don't descend with `"wildcard"` here; instead, just descend with `Nothing`.

Otherwise, you'll have a lot of `wildcard`s mixed into your mangled identifiers, because things like `a ; b` are desugared to `let _ = a ; b`

src/IR/LambdaLift.hs
j-hui: Also, there's no need to use a `let` binding here; just inline it at the `return`

src/IR/LambdaLift.hs
```diff
@@ -150,11 +177,12 @@ liftProgramLambdas p = runLiftFn $ do
...
+liftLambdasTop (v@(VarId lamIdent), lam@(I.Lambda _ _ t)) = do
   let (vs, body) = I.collectLambda lam
       vs'        = zip vs $ fst (collectArrow t)
-  newScope $ catMaybes vs
-  (liftedBody, newTopDefs) <- extractLifted $ liftLambdas body
+  (liftedBody, newTopDefs) <- extractLifted $ descend (Just lamIdent) $ do
+    newScope (catMaybes vs)
+    liftLambdas body
   let liftedLambda = I.makeLambdaChain vs' liftedBody
   return $ newTopDefs ++ [(v, liftedLambda)]
```
j-hui, line 187: You should be able to inline `liftedLambda` here.

src/IR/LambdaLift.hs
```diff
@@ -182,24 +210,26 @@ liftLambdas (I.Prim p exprs t) = do
...
       vs'        = zip vs $ fst (collectArrow t)
-  (liftedLamBody, lamFreeTypes) <-
-    descend $ newScope (catMaybes vs) >> liftLambdas body
+  lamName                       <- getFresh
+  fullName                      <- prependTrail lamName
+  (liftedLamBody, lamFreeTypes) <- descend (Just lamName) $ do
+    newScope (catMaybes vs)
+    liftLambdas body
   let liftedLam = I.makeLambdaChain
         (map (B.first Just) (M.toList lamFreeTypes) ++ vs')
```
j-hui, line 219: nit: `import Data.Bifunctor (first)` instead of using a qualified import, for consistency with other modules

src/IR/LambdaLift.hs
```diff
@@ -209,20 +239,31 @@ liftLambdas (I.Match s arms t) = do
...
 liftLambdasInArm (I.AltLit l, arm) = do
-  (liftedArm, armFrees) <- descend $ liftLambdas arm
+  (liftedArm, armFrees) <- descend Nothing $ liftLambdas arm
   modify $ \st -> st { freeTypes = armFrees }
   return (I.AltLit l, liftedArm)
 liftLambdasInArm (I.AltDefault b, arm) = do
-  (liftedArm, armFrees) <-
-    descend $ F.forM_ b addCurrentScope >> liftLambdas arm
+  (liftedArm, armFrees) <- descend Nothing $ do
+    F.forM_ b addCurrentScope
```
j-hui, line 260: nit: `import Control.Monad (formM_)` instead of using a qualified import, for consistency with other modules

src/IR/LambdaLift.hs
```diff
@@ -24,28 +25,34 @@ import           Control.Monad.State.Lazy       ( MonadState
...
+import           Data.List                      ( intersperse )
 import qualified Data.Map                      as M
 import           Data.Maybe                     ( catMaybes )
 import qualified Data.Set                      as S
 
 -- | Lifting Environment
 data LiftCtx = LiftCtx
   { globalScope  :: S.Set I.VarId
-  {- ^ `globalScope` is a  set containing top-level identifiers. All scopes,
+  {- ^ 'globalScope' is a  set containing top-level identifiers. All scopes,
```
j-hui, line 36: nit: two spaces

j-hui (approved):
Looks great, though I left a bunch of stylistic nits. In particular, you can afford to

Also, could you change the doc comment for [`LiftFn`](https://github.com/ssm-lang/sslang/pull/70/files#diff-59d52e9392d139b5b565076a531d8c6d4680e38697048502a8146fc7415c73d2R60) to use `-- |`, and add module documentation at the top [similar to this](https://github.com/ssm-lang/sslang/blob/9ad72beb4d58944f340a367081740b4a48c243f9/src/IR/LowerAst.hs#L2)

PR #71, opened 2022-01-08
Mangle identifiers for testing equivalence

j-hui:
The following two programs are equivalent:

```
id a : a -> a = a
```

```
foo bar : baz -> baz = bar
```

and yet directly testing for exact equivalence will fail to recognize that these are not just semantically but _structurally_ equivalent.

This PR introduces a lightweight mechanism for telling the tests how not to split hairs over _specific_ identifiers, which may be compiler-generated and thus unstable, yet otherwise semanticaly insignificant.

This PR uses a technique first described in the ["Scrap Your Boilerplate" paper](https://www.microsoft.com/en-us/research/publication/scrap-your-boilerplate-with-class/), which had been sitting open in my browser for a few ~weeks~ months at this point. The technique is implemented in the [syb package](https://hackage.haskell.org/package/syb-0.7.2.1).

This essentially gives us an extremely extensible way of mapping over _any_ user-defined ADT, and map only over specific types embedded in that ADT structure. For example, we can ask it to traverse through any ADT and apply some transformation over all `String` types, or sum together all the `Int`s inside, or monadically mangle all `VarId`s therein.

In other words, it allows us to _generically_ traverse over any ADT---perfect for when we are dealing with large, sprawling ASTs and IRs. The only boilerplate we have to actually do for any data type we define is derive `Typeable` and `Data` instances.

@hmontero1205 and @XijiaoLi this should help you write tests for your passes without worrying about the specific names you generate; would appreciate a review from either or both of you.

In this PR, I'll also clean up the comments and stuff in `Common.Identifiers` and `IR.IR`, since I'm touching those files anyway.

### DIFF
--- a/src/Common/Identifiers.hs
+++ b/src/Common/Identifiers.hs
@@ -3,8 +3,40 @@
 {-# LANGUAGE OverloadedStrings #-}
 {- | The types of identifiers used across the compiler.
 
-These are defined as newtypes (rather than as type aliases) so that they cannot
-be accidentally compared with one another.
+The types 'VarId', 'TVarId', 'DConId', and 'TConId' are used in the IR for data
+variables, type variables, data constructors, and type constructors, which each
+inhabit a separate namespace. The underlying 'Identifier' type is used in the
+AST, before these different kinds of identifiers are differentiated by the
+lowering phase.
+
+These are defined as newtypes (rather than as type aliases) that ultimately
+abstract over the underlying 'String' Identifier. We do this for a few reasons:
+
+    * The particular newtype carries semantic meaning in the IR, e.g., 'VarId'
+      is an identifier for data variables and nothing else. This prevents
+      semantically distinct items from being used in place of one another
+      (without explicit coercion). For example, this will prevent accidentally
+      using a type variable identifier as a data constructor.
+
+    * In the future, this may be extended to carry other (non-semantic)
+      metadata, such as source code location.
+
+    * It allows us to freely attach typeclass instances to this type without
+      FlexibleInstances (since 'String' is a type synonym for @[Char]@).
+
+Users should never need to specifically use the data constructor for each
+newtype, e.g., @VarId (Identifier "foo")@. Instead, each identifier belongs to
+the 'IsString' typeclass, that can be written as @fromString "foo"@, so that the
+appropriate type can be inferred from the context where the identifier is used.
+Furthermore, the 'fromString' call can be automatically inserted by the
+OverloadedStrings GHC extension, meaning we can just write @"foo"@.
+
+All identifier types are instances of the 'Identifiable' typeclass, which allows
+us to write generic functions that operate over any kind of identifier. One can
+explicitly convert from an identifier to another using the handy 'fromId'
+helper; which /specific/ type of identifier may be inferred from the outer
+context, or explicitly annotated (e.g., @fromId i :: VarId@ to construct
+a 'VarId' out of identifier @i@).
 -}
 module Common.Identifiers
   ( Identifiable(..)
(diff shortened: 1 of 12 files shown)


test/trans/Tests/NormalizeSpec.hs
```diff
@@ -0,0 +1,107 @@
+{-# LANGUAGE QuasiQuotes #-}
+-- | Sanity check test suite for equivalent and unequivalent programs.
```
j-hui, line 2: This entire test suite clarifies what it means for programs to be equivalent normalization (i.e., mangling, aka alpha renaming). In the future we may want to introduce further aspects of normalization that can be tested here.

j-hui (commented):
This PR comment should guide reviewers through the most important aspects of this PR.

src/Common/Identifiers.hs
```diff
@@ -49,106 +105,84 @@ instance ToIdent Identifier where
...
-  deriving Identifiable via Identifier
-  deriving Semigroup via Identifier
-  deriving Monoid via Identifier
-  deriving Pretty via Identifier
+  deriving Typeable
+  deriving Data
 
--- | A name to be bound; 'Nothing' represents a wildcard, e.g., @let _ = ...@
+instance Show TVarIdx where
+  show (TVarIdx i) = "'" ++ show i
```
XijiaoLi, line 180: Could we change this string representation to something else? `'0` is pretty nice, but then `brittany` will throw a ParseError when I use it to pretty print the AST, since it cannot handle the single `'` before the index and will consider it to be an unbalanced quotation case. (That's why I changed it to `show (TVarIdx i) = "t" ++ show i` in my type inference PR).

src/Common/Identifiers.hs
```diff
@@ -1,40 +1,96 @@
...
+    * In the future, this may be extended to carry other (non-semantic)
+      metadata, such as source code location.
+
+    * It allows us to freely attach typeclass instances to this type without
+      FlexibleInstances (since 'String' is a type synonym for @[Char]@).
+
+Users should never need to specifically use the data constructor for each
+newtype, e.g., @VarId (Identifier "foo")@. Instead, each identifier belongs to
+the 'IsString' typeclass, that can be written as @fromString "foo"@, so that the
+appropriate type can be inferred from the context where the identifier is used.
```
XijiaoLi, line 30: Neat!

XijiaoLi (approved):
This is really helpful, and the PR is also easy to follow, thanks!

src/Common/Identifiers.hs
j-hui: You raise a good point, but I would actually encourage you to just remove these altogether in your branch, since there's really no need to represent type variables as integers (just makes output harder to read). I kept this in here in case it was still being used elsewhere)

src/Common/Identifiers.hs
```diff
@@ -160,3 +183,33 @@ isCons i | null s    = False
...
+    case M.lookup i idMap of
+      Just i' -> return i'
+      Nothing -> do
+        let ctr' = ctr + 1
+            i'   = fromString $ "mang" <> show ctr'
+        put (ctr', M.insert i i' idMap)
+        return i'
+
+-- | Mangle all type and data variable identifiers.
+mangleVars :: (Data a) => a -> a
```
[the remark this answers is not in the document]
Rewbert: By mangle, you mean rename? I guess at some point down the road when you monomorphise programs, you'll need to rename the constructors anyway.

src/Common/Identifiers.hs
Rewbert: I believe the `OverloadedStrings` language-extension can help you here :) That way you can in most cases drop the `fromString` call yourself.

Normally, a string literal `"x"` gets the inferred type `String`, but with the `OverloadedStrings` extension enabled the inferred type is instead `isString a => a`. GHC will try to solve the constraint for you :)

src/Common/Identifiers.hs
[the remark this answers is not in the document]
j-hui: Mangle and rename mean that same thing (to me)! 

We're actually going to take out the monomorphization pass since the runtime supports true polymorphic code generation.

src/Common/Identifiers.hs
j-hui: I already use that in a few modules!

hmontero1205 (approved):
NICE

PR #76, opened 2022-02-02
Fix doc build CI

j-hui:
the old recipe was missing a build strategy that specifies a resolver and os, and that was hidden by the lack of a hook to run it during a PR.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -28,7 +28,7 @@ jobs:
 
     - name: Build documentation
       run: |
-        stack haddock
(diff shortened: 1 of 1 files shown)


XijiaoLi (approved):
LFGGGGG

j-hui:
Building documentation took 7 minutes!! That's so insanely long. It's because it built documentation for _every dependency_ of sslang as well.

I'm going to specify a flag and see if it speeds things up a little.

j-hui:
TWG [there we go]

PR #82, opened 2022-02-21
Add support for recursion in type inference

XijiaoLi:
Currently we can correctly type check terminated direct recursion and indirect recursion in type inference.


src/IR/HM.hs
```diff
@@ -107,12 +107,24 @@ runInferFn (InferFn m) = evalStateT
 inferProgram :: I.Program Ann.Type -> Compiler.Pass (I.Program Classes.Type)
 inferProgram p = runInferFn $ do
   typeDefs' <- recordADTs $ I.typeDefs p
+  recordFDefs $ I.programDefs p
```
j-hui, line 110: This is only called for the top-level, right? So this doesn't work for nested co-recursive functions?

j-hui (approved):
Sorry for taking so long to review this! I haven't had a very detailed look, but the test cases suggest this works almost as advertised.

After we resolve my comment, let's merge this in and cross potentially broken bridges when we get to them (: 

Another question:

Can this handle types of fields for nested pattern matches? In particular:

```
type MyBool
  MyFalse
  MyTrue

type Nat
  O
  S Nat

myNot b =
  match b
    MyFalse = MyTrue
    MyTrue = MyFalse

isOdd x =
  match x
    O = MyFalse
    S xx = myNot (isOdd xx)
```

XijiaoLi:
> Sorry for taking so long to review this! I haven't had a very detailed look, but the test cases suggest this works almost as advertised.
> 
> After we resolve my comment, let's merge this in and cross potentially broken bridges when we get to them (:
> 
> Another question:
> 
> Can this handle types of fields for nested pattern matches? In particular:
> 
> ```
> type MyBool
>   MyFalse
>   MyTrue
> 
> type Nat
>   O
>   S Nat
> 
> myNot b =
>   match b
>     MyFalse = MyTrue
>     MyTrue = MyFalse
> 
> isOdd x =
>   match x
>     O = MyFalse
>     S xx = myNot (isOdd xx)
> ```

I believe it can!

src/IR/HM.hs
XijiaoLi: Do you mean nested co-recursive functions defined with let-bindings? No, these are not supported at this point.

PR #84, opened 2022-03-01
Pattern match anomaly checker

LeoQiao18:
Based on "Warnings for pattern matching" paper ([link](http://moscova.inria.fr/~maranget/papers/warn/warn.pdf)).

Currently implemented as a separate module that does not affect the rest of the compiler pipeline. Supports exhaustion check and finding useless clause. Unit test written in `match` test module.

Possible further improvements: provide counter-example for non-exhaustion case.

TODO: combine and merge with AST patterns.


j-hui:
Marking as Draft until thus lands changes to the main compiler pipeline

LeoQiao18:
This commit is a very ugly/work-in-progress version of the exhaustion check. I will now work on integration tests for this checker pass. With the tests, I will refactor the code to a more stylistically reasonable state.

PR #85, opened 2022-03-04
Turn unapplied and partially applied adts into regular function application

EmilySillars:
So far, this draft PR includes code changes to generate top-level constructor functions for each data constructor in each ADT definition. 

Next Steps: 
- introduce a reader monad into deconToFunc.hs to hold information on arity of each data constructor (need this information to determine when a data constructor is unapplied/partially applied vs nullary)
- traverse IR to find instances of unapplied and partially applied data constructors, and replace them with calls to constructor functions.


EmilySillars:
Per our discussion earlier this afternoon, I removed special casing of fully applied data constructors from this PR. The current branch as it stands accomplishes the following tasks:
1) creates top-level constructor functions for each data constructor in an ADT definition (excludes nullary data constructors)
2) stores arity of each data constructor in a lookup table inside a reader monad
3) traverses the AST and replaces each data constructors with a call to its corresponding top-level constructor function 

Notable File changes:
- IR.hs: added a function makeAppChain to create a nested application node
- typesystem.hs: added a function peel to discard the outer arrow type of a nested arrow type
- DConToFunc.hs: implementation of formerly mentioned three tasks
- adt-partially-applied.ssl: test file for tasks

Lastly, I didn't replace nullary data constructors with top-level constructor functions. I thought I should wait on this because codegen.hs does not support 0 argument functions right now (if I make this change it will break regression tests that use nullary data constructors).

src/IR/Types/TypeSystem.hs
```diff
@@ -98,6 +98,15 @@ dearrow t = case injectBuiltin t of
   Just (Arrow a b) -> Just (a, b)
   _                -> Nothing
 
+{- | Helper to discard the outer layer of a nested 'Arrow' in any 'TypeSystem'.
+
+If the argument is not a nested Arrow, this function returns Nothing.
+-}
+peel  :: TypeSystem t => t -> Maybe t
```
j-hui, line 105: Isn't this just `snd <$> dearrow`?

src/IR/DConToFunc.hs
```diff
@@ -0,0 +1,166 @@
...
+  (Square, body) 
+  body = fun arg0 { App L R t } : Int -> Shape
+   where L = Square : Int -> Shape
+         R = arg0 : type Int
+         t = Shape, because the type of a fully applied data constructor is its type constructor
+
+  Next consider the top-level func for Rect:
+  (Rect, body*) *Notice the body of the lambda contains another lambda (curried arguments),
+                 and the Application expression contains another Application (curried arguments).  
+  body = fun arg0 { fun arg1 { App L R t } : Int -> Shape } : Int -> Int -> Shape
```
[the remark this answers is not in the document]
EmilySillars: I added a description at the top without IR syntax, but I also kept the old description and put it under a section 'Representing constructor functions in the IR'. I think it might be helpful to keep because it goes over the types of nested lambdas and their arguments. If you think the IR syntax description is too confusing or not helpful, I will remove it. Let me know what you think.

src/IR/DConToFunc.hs
```diff
@@ -0,0 +1,172 @@
...
+  arities (_, TypeDef { variants = vars }) = map (second variantFields) vars
+
+{- | 'dConToFunc' modifies programDefs and traverses the IR to accomplish two tasks:
+
+  1) Add top-level constructor functions for each non-nullary 'DCon' to progamDefs
+  2) Turn non-nullary data constuctors into calls to top level constructor funcs
+-}
+dConToFunc :: I.Program Poly.Type -> Compiler.Pass (I.Program Poly.Type)
+dConToFunc p@I.Program { I.programDefs = defs, I.typeDefs = tDefs } =
+  runArityFn p' (createArityEnv tDefs)
```
j-hui, line 113: This could look nicer if you just swap the order of arguments of `runArityFn`

j-hui: You could just inline `p'`:

```haskell
runArityFn (createArityEnv tDefs) $ do
  ...
```

src/IR/DConToFunc.hs
```diff
@@ -0,0 +1,172 @@
...
+ where
+  func_name = nameFunc dconid -- distinguish func name from fully applied dcon in IR
+  lambda    = I.makeLambdaChain (first Just <$> params) body
+  body      = I.zipApp dcon args
+  dcon      = I.Data (fromId dconid) t
+  args      = zip (reverse $ uncurry I.Var <$> params) (tconTyp : ts)
+  tconTyp   = Poly.TCon tcon []
+  (t : ts) =
+    reverse $ foldl1 arrow . (++ [tconTyp]) <$> tail (inits (snd <$> params))
+  {- ^ Turns a list of types into arrow types representing nested lambdas
```
j-hui, line 155: Does this appear in haddock? If not, remove the `^`. If so, add `@` around code snippets in the comments for better markup.

j-hui (approved):
Looks great, just some small nits. Feel free to merge after addressing those!

src/IR/DConToFunc.hs
EmilySillars: Regardless of the carrot this comment doesn't appear in haddock. Only dContoFunc appears to be documented.
image
Is this normal? The comments are visible in the source:
image

j-hui: Yeah, I guess only documentation for top-level declarations are rendered

PR #86, opened 2022-03-10
Support codegen for closures

j-hui:


j-hui:
Reviewed in-person by @sedwards-lab

PR #88, opened 2022-03-30
Substitute primitives

j-hui:
Still incomplete, but the substitution logic is remarkably easy thanks to SYB-style generic programming.

I still need to tweak the lexer to accept identifiers like `#new` etc. and then fix the scope checker to allow those.

### DIFF
--- a/src/IR/SubstMagic.hs
+++ b/src/IR/SubstMagic.hs
@@ -1,0 +1,44 @@
+-- | Substitute AST nodes with magical primitives.
+{-# LANGUAGE OverloadedStrings #-}
+module IR.SubstMagic
+  ( substMagic
+  ) where
+
+import qualified IR.IR                         as I
+
+import           Data.Data                      ( Proxy(..) )
+import           Data.Generics                  ( Data(..)
+                                                , everywhere
+                                                , mkT
+                                                )
+
+{- | Substitute AST nodes with magical primitives.
+
+Implemented as a syb-style generic tree traversal.
+
+Example usage:
+
+@@
+-- Given:
+myExpr :: Expr Poly.Type
+
+-- Do:
+substMagic (Proxy :: Proxy Poly.Type) myExpr
+
+-- Given:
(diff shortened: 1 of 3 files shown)


src/IR/SubstMagic.hs
```diff
@@ -0,0 +1,44 @@
...
+
+-- Given:
+myProm :: Program Annotated.Type
+
+-- Do:
+substMagic (Proxy :: Proxy Annotated.Type) myProg
+@@
+-}
+substMagic :: (Data t, Data a) => Proxy t -> a -> a
+substMagic p = everywhere $ mkT $ substMagicExpr p
```
yiming-fang, line 36: Is this pass going to be inserted before lowering AST?

j-hui: This pass as it's currently written would perform the substitution on the already-lowered IR.

I'm considering changing strategy and doing this as part of the lowering phase instead.

yiming-fang (approved):

j-hui:
It turns out the lexer is a little too broken to support the approach I'd like to take.

What I wanted to do is to only expose primitives via the `#` form, i.e., `#dup` `#drop`, as well as `#new` and `#deref`. That way, `new` and `deref` could be exposed as regular functions in a prelude that we prepend to the program source text (since we lack any kind of module system):

```
new a = #new a
deref a = #deref a
a + b = #plus a b
a - b = #sub a b
// etc.
```

By declaring these as proper functions, we can also use them in all the same ways (our current approach prevents primitives from being partially applied or passed to higher-order functions, e.g., we can't do `map new [1, 2, 3]`).

However, the lexer/parser seems to choke on just `new` and `deref` being defined in this way.

For the purposes of unblocking @yiming-fang on #78 , I propose merging this in first, to expose the `substMagic` helper to test cases. In the meantime, I'll work on doing this in the more principled way I suggested above, which will also require fixing the lexer (i.e., picking up #38 )

PR #89, opened 2022-04-18
Support stray || inserted due to leading comments

j-hui:
Fix #79

PR #90, opened 2022-04-18
Add nestable block comments

j-hui:
A current problem is that something like `&/*` is being parsed as a single operator, rather than the `&` operator followed by a 

The problem comes from the regexes we use for user-definable operators:

```
$symbol = [\!\#\$\%\&\*\+\-\.\/\<\=\>\?\@\\\^\|\~]
@operator = $symbol ( $symbol | [\_\:\"\'] )*
```

Since both `/` and `*` are symbols, `/*` (and combinations using it) is a valid operator.

Haskell and OCaml sidestep this issue because neither of them allow `{}` or `()` to be used inside of operators.

Proposed resolution to this problem is just switch to Haskell-style comments (which I personally find more elegant). That would also allow symbols such as `//`.

Another simple solution is to drop `/` from the set of supported symbols (but that precludes fun symbols such as `/=` and `//`.

---

This commit should follow #89  and #38


j-hui:
Stephen's arguments against Haskell-style comments: so many languages use C-style comments, we should try to conform to that if possible.

j-hui:
@sedwards-lab some follow-up on the user-defined operator issue:

This is the Regex I came up with:

```
$symbolLeading    = [\!\#\$\%\&\+\-\.\<\=\>\?\@\\\^\|\~]
$symbolAfterSlash = [\!\#\$\%\&\+\-\.\<\=\>\?\@\\\^\|\~\_\:\"\']
$symbolAny        = [\!\#\$\%\&\+\-\.\<\=\>\?\@\\\^\|\~\_\:\"\'\*]

@operator = \/ | ( $symbolLeading | \/ $symbolAfterSlash ) ( $symbolAny | \/ $symbolAfterSlash )* \/?
```

This was constructed using the NFA-to-regex algorithm I first learned about from Sipser's textbook, based on the following GNFA:

```mermaid
stateDiagram-v2
[*] --> FirstSymbol
FirstSymbol --> AfterSlash : /
FirstSymbol --> AnySymbol : $symbolLeading

AfterSlash --> AnySymbol : $symbolAfterSlash


AnySymbol --> AfterSlash : /
AnySymbol --> AnySymbol : $symbolAny

AfterSlash --> [*] : epsilon
AnySymbol --> [*] : epsilon
```

https://en.wikipedia.org/wiki/Maximal_munch

j-hui:
I'm going to go ahead and just merge this one; we can follow up with weird operator/block comment marker conflicts later when we deal with operators for real.

PR #91, opened 2022-04-20
Improve partially applied adt test case

EmilySillars:
When I wrote the test case for partially applied ADTs, I mistakenly through the type inference phase wasn't able to check them. In reality, I simply needed to add parentheses around my expressions. 
Very sorry for this extra merge - after this update we should have more complete test case coverage for partially applied ADTs.


j-hui (approved):
LGTM!

Issue #51, opened 2021-11-17
Prettify pretty printer


j-hui:
The pretty printer output is not pretty---it's time to write an actual pretty printer for the IR.

j-hui:
Should probably follow #38

j-hui:
Note: Emily made a really good start on this in #92, though there are still some issues with precedence.

j-hui:
Completed in #150 and #155

PR #92, opened 2022-04-20
Pretty print

EmilySillars:
Addressing issue #51 
To add indentation/make more specialized formatting for different AST nodes, I added another layer of indirection between the IR and the pretty printer output:
```
src/IR.hs:         indentPretty   

                        |
                        |
                        v

src/IR/IR.hs:           ws       // Whitespace type class

                        |
                        |
                        v

src/IR/IR.hs:         pretty     // Pretty type class
```
The middle layer sort of acts as a more adjustable wrapper for the pretty function. I'm not convinced a typeclass is definitely the right tool here, but I am using that for now and can adjust the code structure as I get a better idea of what is needed.
Currently:
- hello1.ssl can be exported and fed back into the compiler successfully, and looks much more readable.
-  Input hello1.ssl program:
```
main ( cout : & Int ) -> () =
  after 10 , (cout : & Int) <- 72
  wait (cout : & Int)
  after 10 , (cout : & Int) <- 101
  wait (cout : & Int)
  after 10 , (cout : & Int) <- 108
  wait (cout : & Int)
  after 10 , (cout : & Int) <- 108
  wait (cout : & Int)
  after 10 , (cout : & Int) <- 111
  wait (cout : & Int)
  after 10 , (cout : & Int) <- 10
  wait (cout : & Int)
```
- output pretty print using `stack exec -- sslc --dump-ir-final tests/hello1.ssl`
```
main cout : (&Int) -> () = 
 after 10, cout <- 72
 wait cout
 after 10, cout <- 101
 wait cout
 after 10, cout <- 108
 wait cout
 after 10, cout <- 108
 wait cout
 after 10, cout <- 111
 wait cout
 after 10, cout <- 10
 wait cout
```

Next steps:
- prettify match statements with indentation
- make sure nested match, lambdas, and lets are indented properly
- add support for pretty printing type definitions
- solve issue where pretty printer spits out constructor functions starting with "__" which cannot be parsed
- double check annotation on/off switch works as expected and can be fed back in


EmilySillars:
Update:
- match and nested match statements are pretty
- ADT type definitions are pretty
- __functions can be removed from pretty printing via flag
- map.ssl and adt-recursive.ssl are pretty and can be successfully fed back in to compiler ( __funcs OFF)

Next Steps:
- fix issue with nested lambdas and __funcs ON
- make sure type annotated output can be fed back in
- make sure rest of examples look pretty and can be fed back in
- double check nested lambdas look pretty and can be fed back in
- refactor code/use better function naming

Example input/output of pretty printer:
Original:
```
type List
  Cons Int List
  Nil

print_list2 (out: &Int) (l: List) -> () =
  match l
    Cons i l0 =
      after 1, out <- i
      wait out
      after 1, out <- 32
      wait out
      print_list2 out l0
    Nil = ()

print_list (out: &Int) (l: List) -> () =
    after 1, out <- 91
    wait out
    after 1, out <- 32
    wait out

    print_list2 out l

    after 1, out <- 93
    wait out
    after 1, out <- 10
    wait out

map (f: Int -> Int) (l: List) -> List =
  match l
    Cons i l0 = Cons (f i) (map f l0)
    Nil = Nil

add (a: Int) (b: Int) -> Int =
  a + b

main (out: &Int) -> () =
  let l1 = (Cons 104 (Cons 101 (Cons 108 (Cons 108 (Cons 111 Nil)))))
  print_list out l1
  let l2 = map (add 1) l1
  let l3 = map (add 4) l1
  print_list out l1
  print_list out l2
  print_list out l3
```
Output without __funcs:
```
type List 
 Cons Int (List)
 Nil 

print_list2 (out : &Int) (l : (List)) -> () = 
 match l
       Cons i l0 = after 1, out <- i
                   wait out
                   after 1, out <- 32
                   wait out
                   print_list2 out l0
       Nil  = ()

print_list (out : &Int) (l : (List)) -> () = 
 after 1, out <- 91
 wait out
 after 1, out <- 32
 wait out
 print_list2 out l
 after 1, out <- 93
 wait out
 after 1, out <- 10
 wait out

map (f : Int -> Int) (l : (List)) -> (List) = 
 match l
       Cons i l0 = Cons (f i) (map f l0)
       Nil  = Nil

add (a : Int) (b : Int) -> Int = 
 a + b

main (out : &Int) -> () = 
 let l1 = Cons 104 (Cons 101 (Cons 108 (Cons 108 (Cons 111 (Nil)))))
 print_list out l1
 let l2 = map (add 1) l1
 let l3 = map (add 4) l1
 print_list out l1
 print_list out l2
 print_list out l3
```
Output with __funcs:
```
type List 
 Cons Int (List)
 Nil 

print_list2 = fun out fun l match l
      Cons i l0 = after 1, out <- i
                  wait out
                  after 1, out <- 32
                  wait out
                  print_list2 out l0
      Nil  = ()

print_list = fun out fun l after 1, out <- 91
wait out
after 1, out <- 32
wait out
print_list2 out l
after 1, out <- 93
wait out
after 1, out <- 10
wait out

map = fun f fun l match l
      Cons i l0 = __Cons (f i) (map f l0)
      Nil  = Nil

add = fun a fun b a + b

main = fun out let l1 = __Cons 104 (__Cons 101 (__Cons 108 (__Cons 108 (__Cons 111 (Nil)))))
print_list out l1
let l2 = map (add 1) l1
let l3 = map (add 4) l1
print_list out l1
print_list out l2
print_list out l3

__Cons = fun __arg0 fun __arg1 Cons __arg0 __arg1
```

src/IR/IR.hs
```diff
@@ -315,49 +320,72 @@ wellFormed (Prim   p    es   _) = wfPrim p es && all wellFormed es
...
   pretty Program { programDefs = ds } = vsep $ map pretty ds
-  -- TODO: type defs
   -- TODO: how to represent entry point?
 
+-- | pretty print a binder, which is an alias for Maybe Varid
+prettyBinder :: Binder -> Doc ann
+prettyBinder (Just varid) = pretty varid
+prettyBinder _            = pretty '_'
+
 instance Pretty t => Pretty (Expr t) where
```
j-hui, line 357: Could you change this instance to `instance Pretty (Expr ())`? I want the fact that type annotations are being omitted to be visible in the type signature

src/IR/IR.hs
```diff
@@ -383,8 +411,97 @@ instance Pretty PrimOp where
...
+- indentation and line breaks
+- parens where necessary
+Omits 
+- let _ =
+- type annotations
+Reverts
+- curried funcs of one arg back to multiple arg funcs
+-}
+
+class (Pretty a) => FMT a where
```
j-hui, line 433: `FMT` should be stylized as `Fmt`

But also, this looks like what `Pretty` should actually be? Is there a reason to introduce another typeclass?

src/IR/IR.hs
```diff
@@ -383,8 +411,97 @@ instance Pretty PrimOp where
...
+   where
+    letexpr = pretty "let" <+> vsep (map def as) <> line <> fmt b
+    def (Just v , e) = pretty v <+> equals <+> align (fmt e)
+    def (Nothing, e) = pretty '_' <+> equals <+> align (fmt e)
+  fmt (Prim After [d, l, r] _) = ae
+    where ae = pretty "after" <+> fmt d <> comma <+> fmt l <+> larrow <+> fmt r
+  fmt (Prim  Assign [l, r] _) = parens $ fmt l <+> larrow <+> fmt r
+  fmt (Match s      as     _) = pretty "match" <+> pretty s <> line <> arms
+   where
+    arms = vsep (map (indent 6 . arm) as)
```
j-hui, line 465: Where does the hard-coded indentation level of `6` come from?

j-hui (commented):
I left some questions here; let's talk more about this in our meeting later!

EmilySillars:
**Update: Incorporated PR feedback**
- removed FMT typeclass and moved all pretty printing changes to the pretty typeclass
- preserved old pretty printer with type checking in the Lengthy typeclass
- updated spaghetti to call lengthy instead of pretty
- tightened up code in DConToFunc.hs/added more documenation
- replaced magic numbers with indentNo constant defined in Pretty.hs
- updated/improved comments

**Next Steps:**
- I think this PR is ready to merge into main now
- Implementing minimum parentheses algo or selective type annotations could be good work for future PRs
- I think the original pretty printer may have only ever passed unit tests? Maybe we can talk more about this at the meeting.

j-hui (changes requested):
I'd like to disentangle `Pretty` from `Lengthy` (or `Dumpy`) entirely if at all possible. In particular, we will end up with three typeclasses for converting ASTs (or IRs) to Strings (or Docs, and then eventually Strings):

- `Pretty`: Show the prettiest-possible, formatted representation of the AST, with indentation, minimal parens, etc.. This doesn't have to be exactly what the user wrote, but it should more or less look like what a human would write. `Pretty` may end up _losing_ a lot of information (in particular, type information), but that's ok. The point is to produce something readable.
- `Lengthy`/`Dumpy`: Show an ugly but parseable representation of the AST. In particular, it should be able to be parsed by our parser without the lexer's layout algorithm having to insert anything. I'm not sure how useful this will end up being, but it should at least give us a _stable_, machine-parseable representation of the program (i.e., doesn't depend on what way we define the AST/IR data types).
- `Show`: Used to dump out the Haskell ADT definition as `String`. This is used to show every last bit of information that's present in some AST/IR, at the cost of readability. This is mainly going to be useful when the `Pretty` instance fails us. We should always derive this automatically rather than hand-write this, so that it is compatible with automatically-derived `Read` instances (i.e., `read . show === id`).

We shouldn't have any typeclass-level dependencies between these three typeclasses (i.e., there shouldn't be `Pretty t => Lengthy t` or anything like that). One instance may defer to another to avoid code repetition, but that shouldn't be baked into the definition or instance declaration of any typeclass. For example, the `Pretty Literal` instance may just defer to `Lengthy Literal` like this:

```haskell
instance Pretty Literal where
  pretty  = lengthy
```

But only for trivial data types where we know the two are represented exactly the same, independent of readability concerns.

src/IR/IR.hs
```diff
@@ -315,53 +320,100 @@ wellFormed (Prim   p    es   _) = wfPrim p es && all wellFormed es
...
+    <>  line
+ where
+  prettyDCon :: (Pretty t) => (DConId, TypeVariant t) -> Doc ann
+  prettyDCon (dcon, VariantNamed args) =
+    pretty dcon <+> hsep (pretty . snd <$> args)
+  prettyDCon (dcon, VariantUnnamed args) =
+    pretty dcon <+> hsep (pretty <$> args)
+
+
+{- | Pretty prints IR Expr nodes without type annotations
```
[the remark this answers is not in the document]
EmilySillars: Yes, I tried to do that but I was getting a type error:
```
• Illegal instance declaration for ‘Pretty (Expr ())’
    (All instance types must be of the form (T a1 ... an)
     where a1 ... an are *distinct type variables*,
     and each type variable appears at most once in the instance head.
     Use FlexibleInstances if you want to disable this.)
```
I will enable flexible instances and make this pretty instead of fmtExpr.

src/Common/Pretty.hs
```diff
@@ -43,11 +48,25 @@ bar = pretty "|"
...
   where opts = LayoutOptions { layoutPageWidth = Unbounded }
+
+{- | Lengthy Typeclass: "pretty" print the IR; used in conjunction with spaghetti
+
+* Translates from IR to Doc representation in one-to-one fashion
+* No simplifications 
+* No whitespace formatting
+* Type annotates everything
+-}
+class (Pretty a) => Lengthy a where
```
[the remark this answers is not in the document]
EmilySillars: When you say that Lengthy should not depend on Pretty, this means I should be defining instances of Dumpy for literals like Int, right? For example, should I be defining something like this:
`instance Dumpy Int where dumpy = pretty`?
Or should it look like
`instance Dumpy Int where dumpy = unsafeViaShow`?

src/IR/Types/Classes.hs
```diff
@@ -38,8 +38,8 @@ data Scheme = Forall [TVarId] Type
 
 instance Show Scheme where
   show (Forall [] t) = "Forall . " ++ show t
-  show (Forall args t)
-    = "Forall " ++ unwords (map show args) ++ " . " ++ show t
+  show (Forall args t) =
+    "Forall " ++ unwords (map show args) ++ " . " ++ show t
```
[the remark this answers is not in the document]
XijiaoLi: For sure, feel free to change it!

src/Common/Pretty.hs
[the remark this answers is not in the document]
j-hui: I mean the class declaration should look like:

```haskell
class Lengthy a where
  lengthy :: a -> Doc
```

The specific instance may use `pretty` to implement `lengthy` (or vice versa), but that dependency should not be required in the typeclass signature

EmilySillars:
Hi John,
     - I believe I have completely separated Dumpy and Pretty per your request
     - I'm not sure about the change you requested on ast.hs; could you respond to my question on what to do for this part?
     - while updating dumpy to work with the regression tests, I found a bug in the dConToFunc.hs pass which I fixed
     - without pattern desugaring (Leo's pass), all the regression tests work with dumpy, however when I test dumpy with pattern desugaring, map.ssl,map-inc.ssl and adt-partially-applied.ssl all fail. I was trying to resolve this issue this week, but I was getting stuck and thought I better check in with you.
    I realize the semester is over now so maybe I shouldn't be working on this, but I remember you saying you wanted to merge in the pretty-print branch soon and I didn't want to leave this PR in an unfinished state for you. If you'd prefer I stop working on this and you can take over from here, let me know. Otherwise I will keep working on getting dumpy to pass all regression tests. Let me know your thoughts/ what you would like me to do!

src/Common/Pretty.hs
```diff
@@ -9,15 +9,20 @@ module Common.Pretty
...
 
 import           Prettyprinter
 import           Prettyprinter.Render.String
 
--- | @typeAnn t d@ annotates document @d@ with type annotation @t@.
-typeAnn :: Pretty t => t -> Doc ann -> Doc ann
-typeAnn t d = parens $ d <> colon <+> pretty t
+{- | @typeAnn t d@ annotates document @d@ with type annotation @t@.
+
+Only used by spaghetti so constrained by Lengthy instead of Pretty for now.
```
j-hui, line 22: "constrained by Lengthy" -> "constrained by 'Dumpy'"

src/IR.hs
```diff
@@ -43,15 +41,18 @@ data TIType
...
 
+type DupDrop = Bool   -- ^ Flag to run with dup/drop inference
+
 -- | Compiler options for the IR compiler stage.
 data Options = Options
-  { mode   :: Mode
-  , tiType :: TIType
+  { mode    :: Mode
+  , tiType  :: TIType
+  , dupDrop :: DupDrop
```
j-hui, line 50: Just use `Bool` as the type here, no need for the type alias

src/IR/DConToFunc.hs
```diff
@@ -114,9 +114,89 @@ createArityEnv defs = M.fromList $ concatMap arities defs
 dConToFunc :: I.Program Poly.Type -> Compiler.Pass (I.Program Poly.Type)
 dConToFunc p@I.Program { I.programDefs = defs, I.typeDefs = tDefs } =
   runArityFn (createArityEnv tDefs) $ do
-    defs' <- everywhereM (mkM dataToApp) defs
-    return p { I.programDefs = defs' ++ concat (createFuncs <$> tDefs) }
+    defs'' <- defs' -- user defined functions
+    return p { I.programDefs = tDefs' ++ defs'' } -- constructor funcs ++ user funcs
  where
+  tDefs' = concat (createFuncs <$> tDefs)
+  -- ^ top-level constructor functions
```
j-hui, line 121: You should only really use `-- ^` for things like data constructors, where you document the item inline

src/IR/DConToFunc.hs
```diff
@@ -114,9 +114,89 @@ createArityEnv defs = M.fromList $ concatMap arities defs
...
+  @
+  This is okay.
+  Now, given the SSLANG program
+  @
+  type Color
+    White
+    Black
+    RGB Int Int Int
+
+  __RGB (r : Int) (g : Int) (b : Int) -> Color =
```
j-hui, line 157: I would argue that the user should not be able to write this sslang program

src/IR/DConToFunc.hs
```diff
@@ -114,9 +114,89 @@ createArityEnv defs = M.fromList $ concatMap arities defs
...
+  RGB __arg0 __arg1 __arg2
+
+  __RGB (r : Int) (g : Int) (b : Int) -> Color = 
+  RGB r (g+2) b
+    
+  main ( cout : & Int ) -> () = ()
+  @
+  This is not okay, because there are two functions named > __RGB.
+  To prevent this duplicate function case, we search defs for any user defined functions
+  that have the same name as our newly inserted constructor functions, and remove them.
```
j-hui, line 179: This seems like it would actually break programs that define functions that happen to clash with compiler-generated functions. Those functions should not be allowed in the first place.

src/IR/IR.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE FlexibleInstances #-}
```
j-hui, line 2: Which definition requires this?

j-hui (approved):
This looks good as it is. Let's merge this for now, and we can continue tweaking things as we run into problems.

Thanks for the great work @EmilySillars !

src/IR/DConToFunc.hs
EmilySillars: I agree that users should not be able to write functions that clash with compiler-generated functions. I remember we changed the scanner to allow users to begin function names with underscores around April or early May. Starting functions with underscores was the only way I knew of to distinguish compiler generated vs user defined functions. 
1) Would you like me to revert that change to the scanner? The issue here is if we do change the scanner, we can no longer read back in the output of the pretty printer, because dcontofunc generates functions that start with underscores. 
2) Alternatively, maybe dcontofunc could output its constructor functions to a different file that gets linked in. Then the output of the pretty printer could be tested (and fed back in to the front of the compiler), and if any functions with name collisions are detected by dcontofunc in the original source, it could throw an error.

src/IR/IR.hs
EmilySillars: `instance (Pretty Expr ()) where` (line 370) requires this.
Without it, I get the error:
```
• Illegal instance declaration for ‘Pretty (Expr ())’
    (All instance types must be of the form (T a1 ... an)
     where a1 ... an are *distinct type variables*,
     and each type variable appears at most once in the instance head.
     Use FlexibleInstances if you want to disable this.)
• In the instance declaration for ‘Pretty (Expr ())’typecheck
```
It doesn't like that I'm only implementing the Pretty typeclass for `Expr t `when `Expr t`'s type variable is unit.

src/IR/DConToFunc.hs
EmilySillars: **Would you like me to address this issue before merging this PR into main?** I could add a check in `dcontofunc` that throws an error on finding name collisions with user defined functions, but this will break our pretty printer tests. I talked a bit more about this here: (https://github.com/ssm-lang/sslang/pull/92#discussion_r907892990)

src/IR/IR.hs
j-hui: Ah right, got it!

src/IR/DConToFunc.hs
j-hui: Let's fix this later! For now, I want to finalize this PR so that I can make some destructive changes on the IR (:

PR #95, opened 2022-04-27
Setup CI environment with test dependencies for faster CI

j-hui:
CI is _still_ frustratingly slow (#64). It turns out our workflow is recompiling the test dependencies every single time, meaning they aren't being cached properly. I'm using this branch to see if I can speed this up.


j-hui:
I suspect this will be much better when building with cabal

PR #97, opened 2022-04-28
Desugar recursive pattern matching

LeoQiao18:
Desugar recursive pattern matches to non-recursive pattern matches in AST. Algorithm is referenced from Chapter 5 (by Phil Wadler) "Efficient Compilation of Pattern-matching" of "The Implementation of Functional Programming Languages" by Simon Peyton Jones.

### DIFF
--- a/src/Front/Pattern/Anomaly.hs
+++ b/src/Front/Pattern/Anomaly.hs
@@ -23,7 +23,7 @@ import           Control.Monad.Reader           ( MonadReader(..)
                                                 , ReaderT(..)
                                                 , asks
                                                 )
-import           Data.Foldable                  ( find )
+import           Data.List
 import qualified Data.Map                      as M
 import           Data.Maybe                     ( mapMaybe )
 import qualified Data.Set                      as S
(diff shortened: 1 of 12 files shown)


LeoQiao18:
Pattern matching stuff are all implemented now, including:
- inexhaustion check
- useless clause detection
- desugar recursive patterns to non-recursive patterns

All test cases are passing except for "drop inference in pattern matching" in the `DropInference` suite. I don't know why it's failing, but I am inclined to believe that it is some problem with the test case itself. @j-hui

src/Front/Pattern/Anomaly.hs
```diff
@@ -23,11 +23,16 @@ import           Control.Monad.Reader           ( MonadReader(..)
                                                 , ReaderT(..)
                                                 , asks
                                                 )
-import           Data.Foldable                  ( find )
+import           Data.List
```
j-hui, line 26: Please make imports explicit

j-hui (approved):
Just one small nit, aside from that LGTM, :ship: it

PR #100, opened 2022-04-30
Fix gc for effectful lets

yiming-fang:
Stop the GC from trying to dup/drop effectful expressions, creating many extra dup( unit ) or drop( unit ).
For example, nothing should be dupped/dropped from 
`wait led` or `after 10 , led <- 10`.


src/IR/DropInference.hs
```diff
@@ -100,12 +100,15 @@ insertDropExpr :: I.Expr Poly.Type -> InsertFn (I.Expr Poly.Type)
 
 -- | Inserting dup/drops into function application with arg
 insertDropExpr (I.App fun arg typ) = do
-  -- fun' <- insertDropExpr fun
   arg' <- insertDropExpr arg
   return $ I.App fun arg' typ
 
+-- | Skip let bindings that have unit type
```
j-hui, line 106: You should remove the `|` from the comment here. That is used to start haddock documentation for top-level definitions, and should only be used one per definition.

src/IR/DropInference.hs
```diff
@@ -168,7 +172,7 @@ insertDropAlt (I.AltDefault var, exprReturn) exprMatch = do
   return (I.AltDefault var, retExpr')
 
 -- | Inserting dup/drops into AltData arms
--- | TODO: type is currently wrong and is only a placeholder
+-- | TODO: type info is currently unavailable and is only a placeholder
```
j-hui, line 175: Can you explain more about this TODO?

j-hui (approved):
Thanks Yiming! I would double-check the haddock docs generated from the doc comments; make sure to run `stack haddock` and inspect the generated doc site to see that your comments appear as you expect them to.

src/IR/DropInference.hs
yiming-fang: This TODO is addressing the problem that I do not have the types of DCon's fields from upstream.  
For example if a pattern is
`match x`
`MyData a b -> ...`
the types of `a` and `b` are unavailable because the names `a` and `b` are given in a list of VarIds without other information about them.  The current treatment is to use a placeholder for their types, which I believe to cause no problem because it only affects the types of dup/drop expressions (checked during runtime anyways).

PR #104, opened 2022-08-25
Fix dup/drop inference pass

j-hui:


j-hui:
wtf of all things why is _haddock_ breaking

j-hui:
This PR completes work started by Yiming to fix the insertion of ref counting primitives. It turns out, several bigger changes were needed to stick strictly with caller-dup/callee-drop and get things working completely. In particular, we realized that:

- Variables almost always need to be duped whenever they are referenced
- Primitives also need to abide by the callee-drop part of our paradigm, so I insert additional drop calls in the codegen phase to reflect this.
- Intermediate closure values were leaking for a little while. To fix that, we insisted on dropping intermediate values aggressively at the codegen phase.

In the updated compilation pass, pattern matches are the one exception to the dup-every-var rule, because we essentially elided the drop from after the match with the dup from before it.

To make dups and drops easier to insert and reason with, they are now functional rather than sequential; dups now behave as identity functions that have the side effect of incrementing the reference count of its argument, while drops are now scoped, so `drop e x` will evaluate `e`, then drop `x`.

This PR now enables reference counting by default.

Note that there's still an issue with newnew1, flagged in #105

PR #106, opened 2022-08-30
Non-invasive Nix + Nix Flake dev shell support

LeoQiao18:
Stack and Nix have overlapping responsiblities when it comes to package management. Although Stack has the ability to integrate with Nix, I believe that it is not good to change the Stack configurations just to make this play well with Nix on Nix/NixOS systems.

It turns out that Tweag had a post that discusses this, providing a solution that integrates stack and nix in a "smooth and non-invasive" way. For more information, read the [blog](https://www.tweag.io/blog/2022-06-02-haskell-stack-nix-shell/). In order to have more explicit version pinning, this PR is using a version that extends on the idea in the blog and make use of Nix Flake. This is taken from a [section](https://docs.haskellstack.org/en/stable/nix_integration/#supporting-both-nix-and-non-nix-developers) in Stack's documentation.

In the future, the only maintenance needed for this is to sync the GHC version in `flake.nix` with the GHC version for the resolver being used in Stack.

For those on NixOS or systems with Nix that have Flake enabled, you can drop into the nix development shell using either `nix develop` or `nix-shell`. This also allows users to use `nix-direnv`. However, when I tested with `lorri`, it runs into some error... I'll investigate into this later.

Recall that I was having trouble accessing the regression tests when executing `stack test`. This PR fixes that issue as well! So this is important for at least me...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ haddock-out/
 runtests.log
 regression-tests/out
 
+.envrc
+
 # Created by https://www.toptal.com/developers/gitignore/api/haskell
 # Edit at https://www.toptal.com/developers/gitignore?templates=haskell
 
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,43 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1659877975,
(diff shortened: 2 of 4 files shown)


j-hui (approved):
I don't see any change to existing configs, so as far as I'm concerned this LGTM

j-hui:
@LeoQiao18 while you're at it, could you also make a GitHub action that runs the regression test suite using NixOS?

PR #107, opened 2022-08-31
Support all pattern type annotations and check kinds

j-hui:
Follow up to #103, addressing remaining concerns:

- [x] actually unroll pattern annotations rather than discarding them
- [x] typecheck type definitions, i.e., check the kind of each type constructor
- [x] document everything lmao

### DIFF
--- a/src/IR/Types.hs
+++ b/src/IR/Types.hs
@@ -1,3 +1,4 @@
+-- | Wrapper module around other types-related modules and definitions.
 module IR.Types
   ( module IR.Types.Type
   , typecheckProgram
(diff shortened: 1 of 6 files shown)


j-hui:
rip documentation, may lord have mercy on the soul of whoever has to pick up this code (hopefully not me months later, however likely)

j-hui:
Note that there are still some bugs that I've flagged down in #109 

Also there is just no documentation here. Maybe I'll try to write some tomorrow.

LeoQiao18 (approved):

j-hui:
damn I actually documented everything

j-hui:
LFG

PR #108, opened 2022-09-06
Constraint-based type inference

LeoQiao18:
Implement constraint-based type inference.

The type system HM(X) and its type inference process is described in Chapter 10 of Advanced Topics in Types and Programming Languages, called The Essence of ML Type Inference.

The implementation is closely following the OCaml reference implementation, called `mini`. The source code and other information can be found [here](http://cristal.inria.fr/attapl/).


LeoQiao18:
At this point, I have implemented a big chunk of the infrastructure/abstractions needed for constraint generation and solving. Since the inference algorithm (in particular, the unification process) is inherently very mutable, I am making use of the `STT` monad transformer to have strict internal state threading. I manually altered the `union-find` library code to use `STT` instead of `ST` and put it in the `UnionFind.hs` file.

The next step is to finish these abstractions and begin implementing constraint generation.

j-hui:
Since it doesn't seem like this is ready yet, I'm going to make it a draft for now

LeoQiao18:
This new commit begins working on a new version of constraint-based type *elaboration*. It lives inside the `Constraint` module. It references the [inferno](https://gitlab.inria.fr/fpottier/inferno) library, which is related to Francois Pottier's paper, called "Hindley-Milner Elaboration in Applicative Style".

This is still in-progress, but I just want to commit here to save my work so far. I am about half-way done with constraint solving, but there are still a lot more work to do.

The `IR/Types/Constraints` module is now obsolete and will be removed in the future.

LeoQiao18:
The constraint-based type elaboration pipeline is mostly complete now. The only unimplemented thing is type annotation and grouping mutually-recursive bindings.

The next step is to largely refactor the constraint solver code in order to provide a clean interface to the constraint generator. The current constraint generator is using inconvenient combinators, resulting in ugly-looking constraint generation code.

With better interface of the constraint solver, the task of constraint generation should be a lot simpler and become do-able tasks for new type-system group members.

j-hui:
@LeoQiao18 is this superceded by #122? If so, can you close this?

j-hui:
superseded by #122

PR #112, opened 2022-09-23
Document the InsertRefCounting pass

sedwards-lab:
Add documentation on the InsertRefCounting pass so it looks nice in Haddock; add a test case that exercises the pass to verify the examples in the documentation.


j-hui (approved):
Changes LGTM

PR #113, opened 2022-09-23
add switch for spaghetti, update old comment

EmilySillars:
Adds a compiler switch so the user can print IR after type inference with all type annotations visible. This is an ugly/lispy output but it can be helpful for development.


sedwards-lab (approved):
LGTM

PR #115, opened 2022-09-30
Add lists and tuples

EmilySillars:


j-hui:
How is this related to #116 and #119 ?

EmilySillars:
This was the branch I started before giving those tasks to Max and Wei.  
During one of our meetings, I walked them through making a draft pull request and converted this branch to a draft PR as an example. I will delete this branch and close this PR since it was only for an example!

PR #116, opened 2022-10-04
add tuple tests and add tuple syntax in the frontend.

xiaoqiangheye:
add four tests for tuple.
two for the user defined type.
two for the desired syntax. Note that nsyntax tests can not work since tuple syntax haven't been added.

added syntax for tuple in the parser, added tuple expression, fixed some compiler complains in the frontend.


xiaoqiangheye:
I inserted the pair type before anomaly checking. Now the tests of tuples can pass when I remove the type definition of pair in the file. And I can also define tuples using the new syntax (a,b). The pattern matching is currently not supported.

src/IR/LowerAst.hs
```diff
@@ -162,6 +162,14 @@ lowerExpr (A.OpRegion _ _) =
 lowerExpr (A.Match s ps) =
   I.Match <$> lowerExpr s <*> mapM lowerArm ps <*> pure untyped
   where lowerArm (a, e) = (,) <$> lowerPatAlt a <*> lowerExpr e
+lowerExpr (A.Tuple es) =
+  apply_recurse (I.Data "Pair" untyped) <$> mapM lowerExpr es
+
+
+apply_recurse :: I.Expr I.Annotations -> [I.Expr I.Annotations] -> I.Expr I.Annotations
+apply_recurse e [] = e
+apply_recurse e (x:xs) = apply_recurse (I.App e x untyped) xs
```
EmilySillars, line 171: Is it possible to use zipApp from IR.hs instead of defining a new function?

EmilySillars (changes requested):
**Hi Wei,
     This is looking great! Thank you for your hard work!** 
     I added some individual comments to your code and also outlined some general feedback below:

1) Fix missing newlines
Files that need a newline at the end:
regression-tests/tests/tuple1.out
regression-tests/tests/tuple1.ssl
regression-tests/tests/tuple1_nsyntax.out
regression-tests/tests/tuple1_nsyntax.ssl
regression-tests/tests/tuple2.ssl
regression-tests/tests/tuple2_nsyntax.ssl

2) Pull changes from main so your branch is able to merge

3) Resolve Build Warnings,specifically non-exhaustive pattern matching in src/Front/Ast.hs
Allow pretty printer to output Expr of type tuple by adding this line between 220 and 221:
`pretty (Tuple elts) = parens $ hsep (punctuate comma $ map pretty elts)`

4) Fix failing test case
```
###### Testing tuple1_nsyntax
stack exec sslc -- tests/tuple1_nsyntax.ssl > out/tuple1_nsyntax.c
TypeError (ErrorMsg "Unbound variable: DConId (,,)")
###### FAILED
```

Looks like the error is coming from line 75 in src/IR/Types/Inference.hs,
and seems to be caused by the fact that no "Triple" ADT was added to typedefs like in support for 2-tuples.

You can either 
  1. fully add triples as you have done for tuples, so your test case passes,
  2. remove/unstage this test case from your repo so we don't break the regression-test suite upon merging. 
     (You can always add triples later!)

5) In case you haven't already, run hlint/brittany over your code to make sure formatting is correct


- **Let me know if you have any questions on my feedback.**
- **Once you've addressed the individual comments and performed these steps, 
  let's convert this from draft to regular PR and assign John for a final pass!**

src/IR/LowerAst.hs
xiaoqiangheye: find it a bit hard to use that since needs some complex conversions of expression list.

xiaoqiangheye: I move apply_recurse to a where clause so it can be used as a helper function.

EmilySillars:
Hi @j-hui, 
**1) Wei's changes seem to be causing a unit to test to fail.** 
It looks like NormalizeSpec.hs is unhappy because the last type in program defs has type variables a instead of b. Specifically it expects
```
AnnType (TCon -> [TVar b,TVar b])])
```
but gets 
```
 AnnType (TCon -> [TVar a,TVar a])])
```
instead.
**2) Do understand why he's getting this error or how to fix it? 
3) Maybe the type should actually look like `AnnType (TCon -> [TVar a,TVar b])])`, similar to the pair type definitions?**

I added spaces/indentation to the error message below to make it easier to read:
```
Failures:

  test/trans/Tests/NormalizeSpec.hs:81:5: 
  1) Tests.Normalize accepts equivalence under type variable renaming
       expected: Program {programEntry = VarId main, 
                                 cDefs = "", 
                           externDecls = [], 
                           programDefs = [(VarId id,Lambda (Just (VarId a)) 
                                          (Var (VarId a) (Annotations [])) 
                                          (Annotations [AnnArrows [AnnType (TVar _)] (AnnType (TVar _)),
                                                        AnnType (TCon -> [TVar b,TVar b])])
                                          )],
                              typeDefs = [(Pair,TypeDef {variants = [(DConId Pair,VariantUnnamed [TVar a,TVar b])], targs = [a,b]})]}
        but got: Program {programEntry = VarId main, 
                                 cDefs = "", 
                           externDecls = [], 
                           programDefs = [(VarId id,Lambda (Just (VarId a)) 
                                          (Var (VarId a) (Annotations [])) 
                                          (Annotations [AnnArrows [AnnType (TVar _)] (AnnType (TVar _)),
                                                        AnnType (TCon -> [TVar a,TVar a])])
                                          )], 
                              typeDefs = [(Pair,TypeDef {variants = [(DConId Pair,VariantUnnamed [TVar a,TVar b])], targs = [a,b]})]}

  To rerun use: --match "/Tests.Normalize/accepts equivalence under type variable renaming/"

Randomized with seed 1776827942

Finished in 0.0213 seconds
6 examples, 1 failure, 1 pending
```

src/IR/LowerAst.hs
```diff
@@ -162,6 +162,11 @@ lowerExpr (A.OpRegion _ _) =
 lowerExpr (A.Match s ps) =
   I.Match <$> lowerExpr s <*> mapM lowerArm ps <*> pure untyped
   where lowerArm (a, e) = (,) <$> lowerPatAlt a <*> lowerExpr e
+lowerExpr (A.Tuple es) =
+  apply_recurse (I.Data "Pair" untyped) <$> mapM lowerExpr es
```
j-hui, line 166: FYI: this is the wrong data constructor name. You should use the `tupleId` helper to derive the name of the tuple: https://github.com/ssm-lang/sslang/blob/main/src/IR/Types/Type.hs#L224

src/Front/Parser.y
```diff
@@ -314,6 +314,7 @@ exprAtom
   | id                                  { Id $1 }
   | '(' expr ')'                        { $2 }
   | '(' ')'                             { Lit LitEvent }
+  | '(' exprTups ')'                    { Tuple $2}
```
j-hui, line 317: Doesn't this introduce a shift-reduce conflict?

j-hui:
@EmilySillars the reason this test case fails has something to do with the very blunt way I'm doing alpha-renaming when I mangle everything. I'll fix it

src/IR/LowerAst.hs
xiaoqiangheye: If I also change my inserted data type to the name getting from tupleId, it will generate errors from scope "NameError (ErrorMsg "'(,)' should begin with upper case or begin and end with ':'")". I guess it is because naming it "(,)" might not pass the naming check.

j-hui: Ah ok, I should fix that then.

src/Front/Parser.y
```diff
@@ -314,7 +314,7 @@ exprAtom
   | id                                  { Id $1 }
   | '(' expr ')'                        { $2 }
   | '(' ')'                             { Lit LitEvent }
-  | '(' exprTups ')'                    { Tuple $2}
+  | '(' expr ',' exprTups ')'           { Tuple ($2 : $4)  }
```
xiaoqiangheye, line 317: I think this can fix the shift-reduce conflict.

src/Front/Parser.y
EmilySillars: SR conflict should be fixed [here](https://github.com/ssm-lang/sslang/pull/116#discussion_r1008383733)

src/Codegen/Codegen.hs
```diff
@@ -141,9 +141,14 @@ getsTCon f i = do
...
-  Just a <- fmap f . (`dconInfo` i) <$> gets fnTypeInfo
-  return a
+getsDCon f i = 
+  case i of
+    I.DConId "(,)" -> do 
+                   Just a <- fmap f . (`dconInfo` I.DConId "Pair") <$> gets fnTypeInfo
+                   return a
+    _ -> do 
+         Just a <- fmap f . (`dconInfo` i) <$> gets fnTypeInfo
+         return a
```
EmilySillars, line 151: 1) Run the formatter over this code
2) Instead of a case statement, use pattern matching by inserting between lines 143 and 144:
`getsDCon f (I.DConId "(,)") = do ...`
3) after you're done with the edits, run the formatter again :)

src/IR/Types/Inference.hs
```diff
@@ -70,7 +70,13 @@ withBindings bs = U.local
...
 lookupBinding :: Identifiable i => i -> Infer U.Scheme
 lookupBinding x =
-  U.asks (M.lookup (fromId x) . varEnv) >>= maybe (U.throwError unbound) return
+  if fromId x == DConId "(,)"
+  then
+    do 
+    U.asks (M.lookup (fromId (DConId "Pair")) . varEnv) >>= maybe (U.throwError unbound) return
+  else  
+    do
+    U.asks (M.lookup (fromId x) . varEnv) >>= maybe (U.throwError unbound) return
```
EmilySillars, line 79: run formatter over this code

EmilySillars (commented):
Hi Wei, great work!
I added some tiny nit-picky comments, mostly requiring re-running the formatter over your code. We'll make a plan about next steps at your meeting on Monday.

src/IR/Types/Inference.hs
```diff
@@ -70,7 +70,13 @@ withBindings bs = U.local
 -- | Look up the 'U.Scheme' a data identifier in the variable environment.
 lookupBinding :: Identifiable i => i -> Infer U.Scheme
 lookupBinding x =
-  U.asks (M.lookup (fromId x) . varEnv) >>= maybe (U.throwError unbound) return
+  if fromId x == DConId "(,)"
+  then
+    do 
+    U.asks (M.lookup (fromId (DConId "Pair")) . varEnv) >>= maybe (U.throwError unbound) return
```
EmilySillars, line 76: Here is an example of where you can  replace "(,)" and "Pair" with identifiers defined in Identifiers.hs. We can talk more about this at our meeting.

PR #119, opened 2022-10-14
Add list literal syntax

max-acebal:


max-acebal:
Hi I am still getting this error:
`
stack exec sslc -- tests/lists_sample.ssl > out/lists_sample.c
sslc: src/IR/LowerAst.hs:(119,1)-(164,63): Non-exhaustive patterns in function lowerExpr
`
when trying to run my sample test. As we discussed last week I have been reading through the IR to get an idea of how App, and DConId work. I think my next steps should be adding the code for the ListExpr so that it can be generated by the IR in the way we have been discussing on slack, but unsure where that code should even go since is in LowerAst.

EmilySillars:
Hi Max, Thank you for your comment and apologies on the late reply -- 
Restating compiler pipeline here for context: 
`Scanner --> Parser --> Lowering --> Type Inference --> Optimization --> Code Generation --> C code`

**Lowering Stage:** 
The ouput of the parser gives us a representation of our input sslang program as a tree.
Nodes in the tree represents parts of the input program.
The "lowering" pass implemented in LowerAst.hs transforms every AST node into an equivalient IR node.
After performing lowering, we are still representing our input sslang program as a tree in the compiler.
The difference is that after lowering, there are fewer kinds of nodes in our tree, 
and each node is augmented with space for a type, t.

```
// an AST Expr Node     ----- LowerAst.hs ---->     // an IR Expr Node

data Expr                                            data Expr t              
  = Id Identifier                                      = Var VarId t
  | Lit Literal                                        | Data DConId t
  | Apply Expr Expr                                    | Lit Literal t
  | Lambda [Pat] Expr                                  | App (Expr t) (Expr t) t
  | OpRegion Expr OpRegion                             | Let [(Binder, Expr t)] (Expr t) t
  | NoExpr                                             | Lambda Binder (Expr t) t
  | Let [Definition] Expr                              | Match (Expr t) [(Alt, Expr t)] t
  | While Expr Expr                                    | Prim Primitive [Expr t] t
  | Loop Expr                                          deriving (Eq, Show, Typeable, Data, Functor, Foldable, Traversable)
  | Par [Expr]
  | IfElse Expr Expr Expr
  | After Expr Expr Expr
  | Assign Expr Expr
  | Constraint Expr TypAnn
  | Wait [Expr]
  | Seq Expr Expr
  | Break
  | Match Expr [(Pat, Expr)]
  | CQuote String
  | CCall Identifier [Expr]
  | ListExpr [Expr]
  deriving (Eq, Show)
```

The function lowerExpr inside lowerAst.hs has the type signature
`lowerExpr :: A.Expr -> Compiler.Pass (I.Expr I.Annotations)`
which means it transforms something of type A.Expr to into a monadic I.Expr. 
Compiler.Pass is the monad that wraps most all our compiler computations on/transformations 
of our representation of a sslang prorgam.

**Current Compiler Error:**
You are getting a non-exhaustive pattern match warning because the function lowerExpr 
does not have a case for matching on an A.Expr of instance ListExpr.

You need to add a case to LowerExpr that transforms a ListExpr instance of A.Expr into an App instance of I.Expr.

`ListExpr [Expr] ---- needs to turn into ----> App (Expr t) (Expr t) t`

**How to do this???**

**Here is an exercise to get you started thinking about this. 
Write down your answers and comment them on this draft PR before our next meeting.** 
Also comment any questions you have (or slack message me, whichever you prefer) if you get stuck!

  Exercise:
```
  type List
    Cons Int 
    Nil

  main cin cout = 
   let y = (Cons 1 (Cons 2 (Cons 3 Nil)))
   let x = [1, 2, 3]
   ()
```
 
 1) What does y look like as an AST node?
 2) What does x look like as an AST node?

 3) What should y look like as an IR node?
 4) What should x look like as an IR node?

I'm writing up a worked example + solution for a similar exercise, and will post that here as soon as I'm finished with it.

EmilySillars:
Here is an example of a similar exercise with solutions:
[](https://github.com/ssm-lang/sslang/blob/9604ebd88607152fd36c154dcd532b9f43f2bcd9/doc/lang-design/lowering-example.md)

src/Front/Pattern/Desugar.hs
```diff
@@ -90,6 +90,7 @@ desugarExprs = mapM desugarExpr
 desugarExpr :: A.Expr -> DesugarFn A.Expr
 desugarExpr e@(A.Id  _       ) = return e
 desugarExpr e@(A.Lit _       ) = return e
+desugarExpr e@(A.ListExpr _       ) = return e
```
EmilySillars, line 93: pretty sure this should map desugarExpr over elements in the list. I will fix this real quick along with the newlines.

EmilySillars:
Hi @j-hui , 
- The changes in Max's PR here add a `ListExpr` node to the `AST`.
- For the cases of `ListExpr` in `Desugar.hs` and `Anomaly.hs`, should its elements be recursed on, or ignored? **My gut says for each function that takes a `ListExpr`, that function should be called on each of the `ListExpr`'s elements.** Is this correct? If it is, I think a one line change needs to be made in `Anomaly.hs`.
- Any other changes you think need to be made before we merge?

EmilySillars:
I believe Max's `lists_sample2` test case is failing because of a bug main #125

EmilySillars:
Hi Max, 
Great work so far! We're in the home stretch!
1) merge main into your branch
2) once your branch is up to date with main, replace [this line in front.hs](https://github.com/ssm-lang/sslang/blob/main/src/Front.hs#L87) with
`let astTuple = insertTypeDef pairDef astL` to make sure the string and list desugaring gets threaded through.
3) replace lines 19 and 20 desugarLists.hs with
```
    desugarDef (DefFn v bs t e) = DefFn v bs t $ everywhere (mkT desugarExpr) e
    desugarDef (DefPat b e    ) = DefPat b $ everywhere (mkT desugarExpr) e
```
You will likely need to import Data.Generics for this to work. This change allows you to bypass matching on all the different instance of an Expr when you only care about the ListExpr instance - it uses generics from a haskell library called [SYBP](https://web.archive.org/web/20151121203336/http://web.archive.org/web/20080622204226/http://www.cs.vu.nl/boilerplate).

4) At this point your `lists_sample.ssl` test should pass, but doesn't output anything meaningful. Change this test case to print out the contents of your list so that it's outputting something meaningful (instead of nothing). Change the corresponding lists_sample.out file to match your new output. Don't worry about the unescaped string error from `lists_sample2.ssl`. Delete that test or change it so it doesn't use character literals.
5) resolve all your build warnings/ redundant pattern matches etc.

After you complete these steps I will take a final pass over your code for style/ other nits.

src/Front/DesugarLists.hs
```diff
@@ -1,29 +1,55 @@
...
+something of type ListExpr and returns something of type ListExpr, because a ListExpr is not a type.
+It is possible to have a function that takes in 
+something of type Expr and returns something of type Expr!
+Then you can special case on instance of ListExpr.
+I think this is what you wanted to do.
+-}
+
+appidhelper:: [Expr] -> Expr
+appidhelper ([]) = (Id (I.Identifier "Nil"))
+appidhelper ((h:t)) = Apply (Apply (Id (I.Identifier "Cons")) h) (appidhelper t)
```
EmilySillars, line 45: One last thing -- do you see how in these parts you are creating an identifier from a raw string literal every time you talk about the Cons or Nil data constructors? To make the code cleaner, define two special identifiers for the two List data constructors inside `src/Common/Identifiers.hs`. 
Then you can use the identifier name instead of writing `(I.Identifer "Cons")` everywhere.
[I had Wei do this for tuples](https://github.com/ssm-lang/sslang/blob/main/src/Common/Identifiers.hs#L283). 
Instead of writing `(Identifier "Pair")` everywhere, he was able to replace those chunks of code with just `tempTuple`.
Please do the same thing for Cons and Nil data constructors. 
**Follow up if anything I said here is confusing!**

src/Front/DesugarLists.hs
```diff
@@ -1,29 +1,52 @@
...
+something of type ListExpr and returns something of type ListExpr, because a ListExpr is not a type.
+It is possible to have a function that takes in 
+something of type Expr and returns something of type Expr!
+Then you can special case on instance of ListExpr.
+I think this is what you wanted to do.
+-}
+
+appidhelper:: [Expr] -> Expr
+appidhelper ([]) = (Id nil)
+appidhelper ((h:t)) = Apply (Apply (Id cons) h) (appidhelper t)
```
EmilySillars, line 40: I don't think you need parens around `[]` and `(h:t)`. Remove them if you can.

src/Front/DesugarLists.hs
```diff
@@ -1,29 +1,52 @@
...
 -- For ex, (ListExpr [1, 2, 3]) turns into
--- App (App (Id "Cons") (Lit (LitInt 1) )) 
---     (App (App (Id "Cons") (Lit (LitInt 2))) 
+-- App (App (Id "Cons") (Lit (LitInt 1) ))
+--     (App (App (Id "Cons") (Lit (LitInt 2)))
 --          (App (App (Id "Cons") (Lit (LitInt 3))) (id "Nil")))
 desugarExpr :: Expr -> Expr
-desugarExpr e = e -- TODO: actually implement transformation
+desugarExpr (ListExpr es) = appidhelper es
+desugarExpr (ListExpr []) = (Id nil)
```
EmilySillars, line 51: remove this line (build warning due to redundant patterns)

src/Front/DesugarLists.hs
```diff
@@ -1,29 +1,52 @@
...
+ListExpr is a data constructor that returns an instance of type Expr.
+It's not possible to have a function that takes in
+something of type ListExpr and returns something of type ListExpr, because a ListExpr is not a type.
+It is possible to have a function that takes in 
+something of type Expr and returns something of type Expr!
+Then you can special case on instance of ListExpr.
+I think this is what you wanted to do.
+-}
+
+appidhelper:: [Expr] -> Expr
```
EmilySillars, line 38: could you rename this function something shorter? Maybe just call it helper? Also, could you make this a local function by putting it in a `where` clause? For ex., something like 
```
desugarExpr :: Expr -> Expr
desugarExpr (ListExpr es) = appidhelper es
  where helper:: A.ListExpr -> A.ListExpr
        helper [] = (Id nil)
        ...
desugarExpr e = e
```

src/Front/Pattern/Anomaly.hs
```diff
@@ -119,6 +119,7 @@ checkExpr (A.Match e arms) =
   let (ps, es) = unzip arms in checkExpr e >> checkExprs es >> checkPats ps
 checkExpr (A.CQuote _  ) = return ()
 checkExpr (A.CCall _ es) = mapM_ checkExpr es
+checkExpr (A.ListExpr _)    = return ()
```
EmilySillars, line 122: Instead of returning unit here, could you recurse on the elements of the list? Replace the wildcard with a name and map checkExpr over them?

src/Front/Pattern/Desugar.hs
```diff
@@ -341,6 +341,7 @@ substId old new = substExpr
     in  A.Match (substExpr e) (map substArm arms)
   substExpr e@(A.CQuote _  ) = e
   substExpr (  A.CCall s es) = A.CCall s $ map substExpr es
+  substExpr (A.ListExpr es) = A.ListExpr $ map substExpr es
```
EmilySillars, line 344: This is causing redundant pattern matches. I think you can delete this line because you already have this on line 309?

src/Front/Parser.y
```diff
@@ -314,8 +314,18 @@ exprAtom
...
   | '(' expr ',' exprTups ')'           { Tuple ($2 : $4)  }
 
+-- | List Expression.
+exprList
+  :list                                           { $1 }
+  | {- empty list -}                               { [] }
+
+list
+  :exprAtom                                 { [$1] }
+  | exprAtom ',' list                       { $1 : $3 }
```
EmilySillars, line 327: I can't remember if we ever got the formatter working on your vs code. Can you run the formatter over this file? Or at least make the curly braces on lines 322, 323, 326, and 327 align with the curly braces starting on line 318 and above?

EmilySillars (commented):
Great Work, Max!
- I added some final nits.
- Could you also add two more list test cases, one that is for list of lists, and one that is for lists of tuples?
Something like printing out `x`, where `x` is `[[57,56,55],[54]]`
and printing out `y`, where `y` is `[(57,57),(52,53)]`?
You will need to change the definition of List from being integer specific to being polymorphic:
```
type List a
  Cons a (List a)
  Nil
```
For printing the lists of lists, I would add another function
```
 puts2DList cout lst = ...
```
 and for printing a list of tuples, I would add two functions,
```
putsTuple cout tup = ...      // prints out a tuple

putsListOfTuples cout lst = ...
// ^ where this func calls putsTuple for a single elt,
// ^ and calls putsListOfTuples (recurses) on rest of list
```
 Adding these two cases shouldn't take too long. Message me if you get stuck.

EmilySillars (approved):
LGTM!

Issue #111, opened 2022-09-22
Lambda lift breaks recursive let definitions


gschare:
In `regression-tests/tests/hello13.ssl`, the following function for printing a list of characters fails to compile. It complains that the `puts__` variable is unbound, the error presumably originating when it is recursively called.

```
puts_ putc s =
  let puts__ ss =
        match ss
          Cons c s = putc c
                     puts__ s
          Nil = ()
  puts__ s

// TypeError (ErrorMsg "Unbound variable: Var (VarId puts__) (Type [])")
```

We should allow recursive nested functions _a la_ Haskell. This could be achieved by adding the name to the environment at the moment it is declared, like `letrec` bindings in Racket or OCaml, but I'm not sure if would be negative consequences to that. We could also add an explicit `rec` keyword like those languages do, but the result might be aesthetically subpar.

See also #30 -- seems related.

sedwards-lab:
This is definitely a bug since top-level definitions can handle recursion and the local ones should, too. This may be a bug in the type checker. The question is why it's not working for local definitions yet working for top-level definitions. Is the type error being generated before or after lambda-lifting?

j-hui:
I'll look into it

j-hui:
I can't reproduce the `TypeError` shown. In fact, the code snippet shown was originally commented out; if I use that definition it, passes the typecheck, and only fails later on at the Codgen stage: 

```console
$ cat regression-tests/tests/hello13.ssl
putc_ (cout : &Int) c =
  after 1, cout <- c
  wait cout

eof_ (cout : &Int) _ =
  after 1, cout <- 0
  wait cout

type String
  Cons Int String
  Nil

/*
// old workaround definition
puts_ putc s =
  match s
    Cons c ss = putc c
                puts_ putc ss
    Nil = ()
*/

puts_ putc s =
  let puts__ ss =
        match ss
          Cons c s = putc c
                     puts__ s
          Nil = ()
  puts__ s

// Previously commented out, used to throw this error:
// TypeError (ErrorMsg "Unbound variable: Var (VarId puts__) (Type [])")

main cin cout =
  let putc = putc_ cout
  let puts = puts_ putc
  let eof = eof_ cout
  let hello = Cons 72 (Cons 101 (Cons 108 (Cons 108
                (Cons 111 (Cons 49 (Cons 51 (Cons 10 Nil)))))))
  puts hello
  eof ()

$ sslc regression-tests/tests/hello13.ssl
[...]

PR #120, opened 2022-10-17
Add name mangling pass to make VarIds globally unique

j-hui:
This pass adds a name mangling pass that ensures all variable identifiers are
globally unique, while also making sure that identifier names are actually
valid C identifiers.

My mangling strategy is conservative; when picking a new identifier, it appends
ascending numbers until it finds a name that hasn't already been defined.
Though this is somewhat inefficient when there are consistent clashes, it
produces much more readable mangled names.

This pass also introduces a global identifiers map in the `Program` record that
maps those global identifiers to their original names. This can be used for
nicer pretty-printing, or to ensure that newly generated names do not clash
with any other identifiers.

Still left to do:

- [x] Document everything (for some loose definition of "everything")
- [ ] Use the global identifier in passes that generate new names (punted until later)


j-hui:
The lambda lift pass is still broken with respect to #111 . To fix it, I need to add type information to the `varNames` map, which will allow me to simplify the lambda lifting pass to not have to manage local scope anymore.

yiming-fang (approved):
LGTM!

PR #121, opened 2022-10-27
add doc for language design folks

EmilySillars:
I've been sharing some resources with the language design group this semester over slack, but I know it would be better if there was a section in our doc repo they could consult instead of pinned slack messages. 
This PR adds 
- a folder for front end/lang-design related resources, and a file inside with information on the lowering pass
- a general build-test-compile-run cheat sheet to the doc directory, which is relevant to all sslang compiler developers.


j-hui (approved):
LGTM, thanks for writing this @EmilySillars !

PR #122, opened 2022-10-27
Version 2: constraint-based type infernce

LeoQiao18:
This PR is version 2 of the constraint-based type inference. This version is going to look different from the first one, so I am making a new PR.

With the learnings from version 1, version 2 aims to implement constraint-based type inference in a cleaner and more fitting approach for sslang codebase.


LeoQiao18:
The implementation is now complete. There are a couple of problems left that need to be address.

1. Error regression tests: the current error regression tests are hard-coded with error messages, which is not ideal.
2. `polyref1-fail.ssl`: this PR is now incorrectly typechecking this failure case. A bit of work needs to be done to perform the value restriction technique.
3. `TypeInferSpec.hs`: there is something weird going on with mangling inside the type inference unit test, causing correct inferences to fail the comparison tests.
4. `TypeInferSpec.hs`: I believe `typechecks a program with complicated patterns` unit test case is incorrect.

LeoQiao18:
Regarding the 4 things above:
1. error regression tests: @j-hui says that I can just leave it as is for now. Since it is correctly throwing errors, it is ok despite having different error messages.
2. `polyref1-fail.ssl`: I have added value restriction, so this is throwing error as expected. I am doing a little bit more than our previous type-inference, in that if there is a non-value binding, I will not generalize anything in the recursive binding group, which I think is the correct thing to do.
3. `TypeInferSpec.hs`: @j-hui and I will investigate more about the type variable name mangling issue. There should probably be a more sophisticated type variable name mangling function
4. `TypeInferSpec.hs`: the incorrect unit test case is fixed in PR #127

LeoQiao18:
The PR is now ready for review and merge.

j-hui (approved):
ok cool good

(excellent approve)

PR #126, opened 2022-11-03
Allow recursive patterns in the IR

yiming-fang:
This PR adds recursive pattern matching in the IR and also fixes the later passes that are affected by this change.  All passes after type checking will assume that the patterns are non-recursive even though recursive patterns are expressible.

Recursive patterns in the IR are a prerequisite for moving the desugaring and exhaustion-checking to the IR, i.e., after type checking.

### DIFF
--- a/src/Codegen/Codegen.hs
+++ b/src/Codegen/Codegen.hs
@@ -657,7 +657,7 @@ genExpr (I.Match s as t) = do
     withAltScope label (I.AltLit l) m = do
       blk <- m
       return ([citem|case $exp:(genLiteralRaw l):;|], mkBlk label blk)
-    withAltScope label (I.AltDefault b) m = do
+    withAltScope label (I.AltBinder b) m = do
       blk <- withBindings [(b, scrut)] m
       addBinding b scrut
       return ([citem|default:;|], mkBlk label blk)
--- a/src/Codegen/Codegen.hs
+++ b/src/Codegen/Codegen.hs
@@ -651,9 +651,7 @@ genExpr (I.Match s as t) = do
       destruct <- getsDCon dconDestruct dcon
       cas      <- getsDCon dconCase dcon
       let fieldBinds =
-            zipWith (\field i -> (getAltDefault field, destruct i scrut)) fields [0 ..]
-            where getAltDefault (I.AltDefault b) = b
-                  getAltDefault _ = Nothing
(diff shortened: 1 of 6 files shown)


src/Codegen/Codegen.hs
```diff
@@ -651,7 +651,9 @@ genExpr (I.Match s as t) = do
       destruct <- getsDCon dconDestruct dcon
       cas      <- getsDCon dconCase dcon
       let fieldBinds =
-            zipWith (\field i -> (field, destruct i scrut)) fields [0 ..]
+            zipWith (\field i -> (getAltDefault field, destruct i scrut)) fields [0 ..]
+            where getAltDefault (I.AltDefault b) = b
+                  getAltDefault _ = Nothing
```
LeoQiao18, line 656: Things like this should not be returning `Nothing`. In fact, they should be a compiler error/bug.

@j-hui I think we can just call the `error` function here?

LeoQiao18: We can maybe create a function called `assumeAltDefault` in `IR.hs`, and use that function in the IR passes that assumes non-recursive patterns. This way encapsulates the `error` logic in that one place.

j-hui: Yeah, agree with @LeoQiao18 here

LeoQiao18:
@j-hui I am not super familiar with the `PatDef` and `PatFn`; in particular, how are the patterns in these data types actually being used by the AST? Are the patterns used in definitions like the following?

```haskell
foo [] = ...
foo (x:xs) = ...
```

Or are these patterns something else entirely.

src/IR/IR.hs
```diff
@@ -222,7 +223,7 @@ data Expr t
 
 -- | An alternative in a pattern-match.
 data Alt
-  = AltData DConId [Binder]
+  = AltData DConId [Alt]
   -- ^ @AltData d vs@ matches data constructor @d@, and names dcon members @vs@.
   | AltLit Literal
   -- ^ @AltLit l@ matches against literal @l@, producing expression @e@.
```
j-hui, line 229: Can you change `AltDefault` to be `AltBinder` instead? Because "default" no longer makes sense here.

j-hui (approved):
Awesome work! Just a couple of small nits, but it looks good to me otherwise. Feel free to merge once you've addressed those nits.

PR #127, opened 2022-11-08
Fix incorrect unit test case

LeoQiao18:
Incorrect type inference unit test case. There are quite a bit of style changes from brittany as well.

Interestingly, the test case still passes even after the change.

### DIFF
--- a/test/semant/Tests/TypeInferSpec.hs
+++ b/test/semant/Tests/TypeInferSpec.hs
@@ -9,7 +9,7 @@ import qualified IR
 import qualified IR.IR                         as I
 
 import           Control.Monad                  ( (>=>) )
-import           Debug.Trace                    ( traceM )
+import Debug.Trace (traceM)
 
 infer :: HasCallStack => String -> Pass (I.Program I.Type)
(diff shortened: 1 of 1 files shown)


j-hui:
Oh weird, why is Brittany mangling the style so heavily? It shouldn’t do that.

Can you fix the commit to only fix the test case? Let’s fix the style/figure out what’s going on in another PR.

j-hui (changes requested):
Revert style changes; we can address that in another PR

LeoQiao18:
@j-hui Juding from what is actually being styled, the original code did not look actually styled. For example, the import statement was not even aligned.

j-hui:
The traceM import statement shouldn't even be there; that was left there by mistake.

I remember now; I deliberately omitted styling because it messes with the way I'm using quasiquote blocks with infix functions. I'd prefer we leave it that way for now until we figure out a good consistent style for all test cases (which may involve a little bit of refactoring, or at least shortening of the strings documenting each test case).

LeoQiao18:
Reverted the style changes.

j-hui (approved):
Thanks Leo! Please squash and merge!

PR #128, opened 2022-11-11
Desugar topDef

hzhou07:
With this PR, the top level function definitions with pattern matching will be transformed into a equivalent one starting with a match in function body

> - Top level function definition with no pattern:
> 
> ```
> f  = expr
> ```
> 
> will remain unchanged

> - Top level function definition with one pattern:
> 
> ```
> f a = expr
> ```
> 
> will be desugered to:
> 
> ```
> f x = match x with
>     a -> expr
> ```

> - Top level function definition with no less than two patterns:
> 
> ```
> f (x1,x2) (y1,y2) = expr
> ```
> 
> will be desugered to:
> 
> ```
> f v1 v2 = match (v1, v2) with
>     ((x1,x2), (y1,y2))-> expr
> ```

~~Note: Should fail building until pull[#116](https://github.com/ssm-lang/sslang/pull/116) is merged~~

### DIFF
--- a/src/Front/Pattern/Desugar.hs
+++ b/src/Front/Pattern/Desugar.hs
@@ -85,10 +85,14 @@ desugarDefs = mapM desugarDef
 -- WARN: fail building until tuple implementation is merged
 desugarDef :: A.Definition -> DesugarFn A.Definition
 desugarDef (A.DefFn i ps t e) = do
+  psI <- mapM genNewId ps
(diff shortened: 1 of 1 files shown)


j-hui:
I'm going to close this, since it works on `src/Front/Pattern/Desugar.hs`, and that is now moved to the back-end. This desugaring should be done elsewhere.

PR #129, opened 2022-11-18
Desugar pattern matching in IR

yiming-fang:
This PR will migrate the previous work done by @LeoQiao18 on desugaring pattern matching from AST to IR.

The rationale behind this migration is to ensure that the desugaring pass only happens after type-checking.  Type correctness is assumed by most desugaring algorithms, and doing desugaring in the AST can lead to potential hazards, such as accidentally transforming ill-typed expressions to well-typed expressions or vice versa.


src/IR/DesugarPattern.hs
```diff
@@ -157,7 +157,16 @@ desugarExpr (I.Match e arms t) = do
         )
         arms
     let armsForDesugar = map (first (: [])) arms'
-    desugarMatch [e] armsForDesugar (I.NoExpr t)
+    case e of
+        I.Var _ _ -> desugarMatch [e] armsForDesugar (I.NoExpr t)   -- TODO: add let alias
+        _         -> do
+            v <- freshVar
+            singleLet (I.VarId v) e
+                <$> desugarMatch [e] armsForDesugar (I.NoExpr t)
```
yiming-fang: I think a modification should be made here, please see the new commits.

`<$> desugarMatch [e] armsForDesugar (I.NoExpr t)` 

Should be

`<$> desugarMatch [I.Var var vt] armsForDesugar (I.NoExpr t)`

src/IR/DesugarPattern.hs
```diff
@@ -241,14 +250,21 @@ desugarMatchCons (u : us) qs def = do
...
+                return
+                    ( I.AltData dcon (map makeBinder us')
+                    , I.NoExpr (I.extract def)
+                    )
+            _ -> do
+                body <- desugarMatch
+                    (us' ++ us)
+                    [ (as' ++ as, e) | ((I.AltData _ as') : as, e) <- qs' ]
+                    def
+                return (I.AltData dcon (map makeBinder us'), body)
```
yiming-fang: I think this change caused some problems, because it sets the DCon that are not in the pats directly to NoExpr, which is not desired

src/IR/DesugarPattern.hs
hzhou07: agreed.

PR #118, opened 2022-10-14
Add String Literals

as0501:
I'm currently working on editing the parser to handle our desired syntax for string literals. I made a new test file called hello-new-string.ssl that will print "Hello". I also edited the Scanner.x file to update the string literals action and import the reverse Kleisli arrow from Control.Monad.

PR #131, opened 2022-11-21
Make AST implement typeable and data

EmilySillars:
This change will allow Max (on #119) and Anjali (on #118) to use `everywhere` over the AST, simplifying their passes .


j-hui (approved):
LGTM

PR #132, opened 2022-11-30
add simple inlining pass

EmilySillars:


EmilySillars:
This PR adds a simple inlining pass to the SSLANG compiler. This "simplifier" pass is split into parts, occurrence analysis, followed by inlining.
**Our inliner supports pre-inline unconditionally and post-inline unconditionally,** 
which correspond to replacing binder x with value e when x gets marked by the occurrence analyzer as one of three cases:
1) Dead: Does not appear at all
2) OnceSafe: Appears once, NOT inside a lambda
3) MultiSafe: The binder occurs at most ONCE in each of several distinct case branches; NONE of these ocurrences is inside a lambda

**Regression Tests Added:**
inlining-ex-1.ssl
inlining-ex-2.ssl
inlining-ex-3.ssl
inlining-ex-4.ssl
inlining-ex-5.ssl

 **TODO/Future Work:**
 - Incorporate Constructor Argument Invariant (left as a TODO)
 - Add ability to inline binders introduced in match arms (left as a TODO)
 - Add Callsite Inline phase to the inliner to allow for handling binders marked by OA as one of these cases:
1) OnceUnsafe: Binder occurs exactly once, but inside a lambda.
2) MultiUnsafe: Binder may occur many times, including inside lambdas.
3) LoopBreaker: Chosen to break dependency between mutually recursive defintions.
- Update our substitution set type to only allow for flat substitutions (currently capable of more than a flat map, and doesn't need that capability)

src/IR/Simplify.hs
```diff
@@ -0,0 +1,459 @@
...
+
+-- | Update occInfo for the binder since we just spotted it
+updateOccVar :: I.VarId -> SimplFn ()
+updateOccVar binder = do
+  m <- gets occInfo
+  insidel <- insideLambda
+  insidem <- insideMatch
+  let m' = case M.lookup binder m of
+        Nothing ->
+          error
```
j-hui, line 135: Is this an `error` in the sense that this code should never be reached?

src/IR/Simplify.hs
```diff
@@ -0,0 +1,459 @@
...
+
+-- Simplify Variable Expression
+simplExpr _ ins var@(I.Var v _) cont = do
+  m <- gets subst
+  case M.lookup v m of
+    Nothing -> pure var -- callsite inline, future work
+    Just (SuspEx e s) -> simplExpr s ins e cont
+    Just (DoneEx e) -> simplExpr M.empty ins e cont
+
+-- Simplify Let Expressions
```
j-hui, line 284: Can you convert these into haddock comments?

src/IR/Simplify.hs
```diff
@@ -0,0 +1,459 @@
...
+- App (Expr t) (Expr t) t               DONE
+- Let [(Binder, Expr t)] (Expr t) t     DONE
+- Lambda Binder (Expr t) t              DONE
+- Match (Expr t) [(Alt, Expr t)] t      DONE (TODO: analyze patterns / LHS of arms?)
+- Prim Primitive [Expr t] t             DONE
+-}
+occAnalExpr :: I.Expr I.Type -> SimplFn (I.Expr I.Type, String)
+-- Occurrence Analysis over Let Expression
+occAnalExpr l@(I.Let nameValPairs body _) = do
+  mapM_
```
j-hui, line 382: You can use `forM_` here, which is just defined as follows:

```haskell
forM_ = flip mapM_
```

Then you don't need to wrap the lambda around parens, you can just do

```haskell
forM_ nameValPairs $ \(binder, rhs) -> do $ -- etc.
```

j-hui: (you'll also need to import `forM` from the standard library)

src/IR/Simplify.hs
```diff
@@ -0,0 +1,459 @@
...
+  let m' = case M.lookup binder m of
+        Nothing ->
+          error
+            ( "UDPATE: We should already know about this binder "
+                ++ show binder
+                ++ " :"
+                ++ show m
+                ++ "!"
+            )
+        Just Dead -> do
```
j-hui, line 142: This is not monadic, so it should not have a `do` block

src/IR/Simplify.hs
```diff
@@ -0,0 +1,459 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{-# HLINT ignore "Use record patterns" #-}
```
j-hui, line 6: Hey hey record patterns are pretty nice (:

src/IR/Simplify.hs
```diff
@@ -0,0 +1,459 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
```
j-hui, line 4: What's this for?

j-hui (approved):
Overall this LGTM. There are some small nits, but aside from those, this seems ready to be merged in

src/IR/Simplify.hs
EmilySillars: We didn't need any of them except DerivingVia; thanks!

PR #133, opened 2022-12-08
Tuple changes

xiaoqiangheye:
Add function signature matching on tuple.
Fix the loweringPatAlt to avoid later passes changes.


src/Front/Pattern/Desugar.hs
```diff
@@ -79,11 +79,27 @@ desugarTopDefs = mapM desugarTopDef
 desugarDefs :: [A.Definition] -> DesugarFn [A.Definition]
 desugarDefs = mapM desugarDef
 
--- WARN: only body is desugared
 desugarDef :: A.Definition -> DesugarFn A.Definition
-desugarDef (A.DefFn i ps t e) = A.DefFn i ps t <$> desugarExpr e
+desugarDef (A.DefFn i [] t e) = return (A.DefFn i [] t e)
+desugarDef (A.DefFn i ps t e) =
+  let (pats, rese) = helper ps e (0::Int) in
```
EmilySillars, line 85: I would put `(pats, rese) = helper ps e (0::Int)` inside the where clause and get rid of this let.

src/Front/Pattern/Desugar.hs
```diff
@@ -79,11 +79,27 @@ desugarTopDefs = mapM desugarTopDef
...
+    A.DefFn i pats t <$> desugarExpr rese
+  where helper [] ex _ = ([],ex)
+        helper (p:rps) ex n = 
+         case p of
+           A.PatTup _ -> let (pats, rese) = helper rps (A.Match (A.Id (Identifier ("_temp_id_" ++ show n))) [(p,ex)]) (n+1) in
+                         (A.PatId (Identifier ("_temp_id_" ++ show n)) : pats, rese)
+           _ -> let (pats, rese) = helper rps e (n+1) in
+                (p:pats, rese)
```
EmilySillars, line 93: should be able to rewrite this let as` first (p:) $ helper rps e (n + 1) `

src/Front/Pattern/Desugar.hs
```diff
@@ -79,11 +79,27 @@ desugarTopDefs = mapM desugarTopDef
...
 desugarDefs = mapM desugarDef
 
--- WARN: only body is desugared
 desugarDef :: A.Definition -> DesugarFn A.Definition
-desugarDef (A.DefFn i ps t e) = A.DefFn i ps t <$> desugarExpr e
+desugarDef (A.DefFn i [] t e) = return (A.DefFn i [] t e)
+desugarDef (A.DefFn i ps t e) =
+  let (pats, rese) = helper ps e (0::Int) in
+    A.DefFn i pats t <$> desugarExpr rese
+  where helper [] ex _ = ([],ex)
```
EmilySillars, line 87: what does this helper do? Add comment explaining, and give it a more descriptive name.

EmilySillars, line 87: Are you able to get rid of this case since `ps` is never empty, or do you get  warning?

EmilySillars (commented):
Hi Wei, thank you for your hard work!
1) I'm getting this warning when I build:
```
/home/spinel/sslang/src/Front/Scanner.x:32:1: warning: [-Wunused-imports]
    The import of ‘<=<’ from module ‘Control.Monad’ is redundant
   |
32 | import Control.Monad (when, (<=<))
```
**I realize you didn't originally modify the scanner, but could you go ahead and get rid of this redundant import?**

2) I think instead of inserting the tuple types inside of `Front.hs`, you should create another file called `tuple-lib.ssl `or something that holds all your tuple definitions written in SSLANG. Then copy the contents of that file above all your test cases involving tuples. This is better because tuples should be part of the standard library anyway, and storing them as AST nodes is more fragile than storing them as source code.

3) This unit test failing in semant actually should be failing:
```
  1) Tests.Scoping, scoping redefinitions and shadowing, rejects overlapping identifiers in the same argument pattern
       Did not encounter expected error
       expected: ScopeError (ErrorMsg "x is redefined")
        but got: ()

```
**Could you edit Scope.hs to account for the case when a tuple pattern has two ids named the same thing at the same level? I.E. `f (x,x) = x + x //is wrong`**

**4) Does this PR support 4-tuples? Right now 4-tuples fail.
5) resolve small comments I left below**

EmilySillars (approved):
LGTM!

PR #134, opened 2022-12-11
Add exception

hzhou07:
Exception is introduced in IR to support dead branch for our desugaring algorithm migrated to IR.

Current implementation is using [`throw`](https://github.com/ssm-lang/sslang/blob/main/src/Codegen/LibSSM.hs#L99) with [`INTERNAL_ERROR`](https://github.com/ssm-lang/sslang/blob/main/src/Codegen/LibSSM.hs#L85)

### DIFF
--- a/src/Codegen/Codegen.hs
+++ b/src/Codegen/Codegen.hs
@@ -141,9 +141,9 @@ getsTCon f i = do
 
 -- | Lookup some information associated with a data constructor.
 getsDCon :: (DConInfo -> a) -> I.DConId -> GenFn a
-getsDCon f i = do 
-              Just a <- fmap f . (`dconInfo` i) <$> gets fnTypeInfo
-              return a
+getsDCon f i = do
+  Just a <- fmap f . (`dconInfo` i) <$> gets fnTypeInfo
+  return a
 
 -- | Read and increment the number of cases in a procedure, i.e., @fnCases++@.
 nextCase :: GenFn Int
@@ -651,8 +651,10 @@ genExpr (I.Match s as t) = do
     withAltScope label (I.AltData dcon fields) m = do
       destruct <- getsDCon dconDestruct dcon
       cas      <- getsDCon dconCase dcon
-      let fieldBinds =
-            zipWith (\field i -> (I.getAltDefault field, destruct i scrut)) fields [0 ..]
+      let fieldBinds = zipWith
+            (\field i -> (I.getAltDefault field, destruct i scrut))
+            fields
+            [0 ..]
       blk <- withBindings fieldBinds m
       return ([citem|case $exp:cas:;|], mkBlk label blk)
     withAltScope label (I.AltLit l) m = do
--- a/src/Codegen/Codegen.hs
+++ b/src/Codegen/Codegen.hs
@@ -667,7 +667,9 @@ genExpr (I.Match s as t) = do
   return (val, sStms ++ assignScrut ++ switch cases ++ blks ++ joinStm)
 genExpr I.Lambda{}        = fail "Cannot handle lambdas"
 genExpr (I.Prim p es t  ) = genPrim p es t
-genExpr (I.Exception _ _) = error "cannot happen" -- for me TODO
+genExpr (I.Exception _ t) = do
+  tmp <- genTmp t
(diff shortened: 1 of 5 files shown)


src/IR/IR.hs
```diff
@@ -488,13 +488,12 @@ instance Pretty (Expr ()) where
...
   pretty (Prim (FfiCall s) es _) = pretty s <+> hsep (map (parens . pretty) es)
-  pretty (Prim (CQuote s) [] _) = pretty "$$" <> pretty s <> pretty "$$"
+  pretty (Prim (CQuote  s) [] _) = pretty "$$" <> pretty s <> pretty "$$"
 
   -- pretty (Prim Return [e] _        ) = pretty "return" <+> braces (pretty e)
   pretty (Prim p _ _) = error "Primitive expression not well-formed: " $ show p
-  pretty (Exception et _) = pretty "exception" <+> prettyet et
-    where 
-      prettyet (ExceptDefault l) = pretty l
+  pretty (Exception et _       ) = pretty "exception" <+> prettyet et
```
j-hui, line 495: Change to `"__exception"` to make it clear this is a primitve

src/Codegen/Codegen.hs
```diff
@@ -669,8 +670,9 @@ genExpr (I.Match s as t) = do
 
   (cases, blks) <- bimap concat concat . unzip <$> mapM genArm as
   return (val, sStms ++ assignScrut ++ switch cases ++ blks ++ joinStm)
-genExpr I.Lambda{}      = fail "Cannot handle lambdas"
-genExpr (I.Prim p es t) = genPrim p es t
+genExpr I.Lambda{}        = fail "Cannot handle lambdas"
+genExpr (I.Prim p es t  ) = genPrim p es t
+genExpr (I.Exception _ _) = error "cannot happen" -- for me TODO
```
j-hui, line 675: For now, can you generate an `ssm_throw()`? In particular, use [`throw`](https://github.com/ssm-lang/sslang/blob/main/src/Codegen/LibSSM.hs#L99) with [`INTERNAL_ERROR`](https://github.com/ssm-lang/sslang/blob/main/src/Codegen/LibSSM.hs#L85)

j-hui (commented):
Small nit, which is to change how exceptions are pretty-printed

For the exception implementation in codegen, I recommend using ssm_throw

src/IR/IR.hs
hzhou07: Done.

src/Codegen/Codegen.hs
```diff
@@ -664,8 +667,11 @@ genExpr (I.Match s as t) = do
 
   (cases, blks) <- bimap concat concat . unzip <$> mapM genArm as
   return (val, sStms ++ assignScrut ++ switch cases ++ blks ++ joinStm)
-genExpr I.Lambda{}      = fail "Cannot handle lambdas"
-genExpr (I.Prim p es t) = genPrim p es t
+genExpr I.Lambda{}        = fail "Cannot handle lambdas"
+genExpr (I.Prim p es t  ) = genPrim p es t
+genExpr (I.Exception _ t) = do
+  tmp <- genTmp t
+  return (tmp, [citems|$exp:(throw INTERNAL_ERROR);|]) -- unit instead of temp?
```
j-hui, line 674: Yeah, I think you can just return `unit` here instead of a temp variable

src/IR/IR.hs
```diff
@@ -231,21 +234,26 @@ data Alt
   -- ^ @AltBinder v@ matches anything, and bound to name @v@.
   deriving (Eq, Show, Typeable, Data)
 
+newtype ExceptType
```
j-hui, line 237: Is there a reason you have `ExceptDefault` wrapping a literal here?

j-hui (approved):
I'm happy enough with this as is, so you should merge in, but please also address my question and comment before you do so.

src/IR/IR.hs
hzhou07: I have this at the very beginning. I was thinking about later we might want to let user throw exception, so a literal can be used to distinct different type of Exception Error. That's why I make this for future implementation. But I remember you mentioned sslang users would not throw any exception. Maybe I can need to remove this completely?

src/Codegen/Codegen.hs
hzhou07: Done

PR #136, opened 2022-12-13
Tuple func signature

xiaoqiangheye:
add branch for function signatures


EmilySillars:
I am pretty sure this work is superseded by  #129

j-hui:
@xiaoqiangheye @yiming-fang can you please follow up and determine whether this PR is superseded by #129? My understanding is that it isn't: #129 deals with the expansion of nested patterns at the IR level, whereas this PR only works with the front-end.

In any case, I think we need to clarify what this PR intends to do, since the title implies it is related to function signatures rather than patterns.

xiaoqiangheye:
this will expand a match on function signatures to a match expression. I am
not sure if #129 will does this at IR level.

yiming-fang:
This PR does not really overlap with #129 which is exclusively on the IR, but might be relevant for #128.  I remember when @hzhou07 and I were working on #128 last semester, we did not have PatTup support. Is this what you are adding in this current PR @xiaoqiangheye?

PR #139, opened 2023-02-20
Matching tuples on let expression and top function definitions.

xiaoqiangheye:


src/Front/DesugarPatTup.hs
```diff
@@ -0,0 +1,35 @@
...
+    where desugarTop (A.TopDef defs) = A.TopDef $ desugarDef defs
+          desugarTop t = t
+          desugarDef (A.DefFn i [] t e) = A.DefFn i [] t $ desugarExpr e
+          desugarDef (A.DefFn i ps t e) =
+                let (pats, rese) = desugarPatTup ps e (0::Int) in
+                    A.DefFn i pats t $ desugarExpr rese
+                where desugarPatTup [] ex _ = ([],ex)
+                      desugarPatTup (p:rps) ex n = 
+                        case p of
+                            A.PatTup _ -> let (pats, rese) = desugarPatTup rps (A.Match (A.Id (Identifier ("_temp_id_" ++ show n))) [(p,ex)]) (n+1) in
```
j-hui, line 19: Nit: can you shorten this line, or break it up into multiple? It's very long. Also, you should use `fromString` rather than manually using the `Identifier` constructor; so you should be able to do:

```haskell
(fromString $  "__temp_id_" <> show n)
```

src/Front/DesugarPatTup.hs
```diff
@@ -0,0 +1,35 @@
...
+desugarTupPat :: A.Program -> Compiler.Pass A.Program
+desugarTupPat (A.Program decls) = return $ A.Program $ desugarTop <$> decls
+    where desugarTop (A.TopDef defs) = A.TopDef $ desugarDef defs
+          desugarTop t = t
+          desugarDef (A.DefFn i [] t e) = A.DefFn i [] t $ desugarExpr e
+          desugarDef (A.DefFn i ps t e) =
+                let (pats, rese) = desugarPatTup ps e (0::Int) in
+                    A.DefFn i pats t $ desugarExpr rese
+                where desugarPatTup [] ex _ = ([],ex)
+                      desugarPatTup (p:rps) ex n =
```
j-hui, line 17: Nit: Try not to mix `where` and `let` clauses in the same expression; stick to one or the other.

src/Front/DesugarPatTup.hs
```diff
@@ -0,0 +1,35 @@
+module Front.DesugarPatTup where
+
+import qualified Common.Compiler as Compiler
+import Common.Identifiers
+import qualified Front.Ast as A
+import           Data.Bifunctor                 ( first )
+
+desugarTupPat :: A.Program -> Compiler.Pass A.Program
+desugarTupPat (A.Program decls) = return $ A.Program $ desugarTop <$> decls
+    where desugarTop (A.TopDef defs) = A.TopDef $ desugarDef defs
```
j-hui, line 10: nit: can you move `desugarTop` etc. to the top-level, rather than sticking them all in this huge `where` block?

j-hui: Of course you'll need to add type annotations etc. but that should only serve to better document this code.

j-hui (commented):
Thanks for working on this! I think this code could be made simpler using a generic traversal, i.e., using `everwhere`. Then you don't have to worry about all this recursion.

An example of where I do this is in https://github.com/ssm-lang/sslang/blob/main/src/IR/SubstMagic.hs

src/Front/DesugarPatTup.hs
```diff
@@ -0,0 +1,74 @@
...
+--                         (A.DefPat p@(A.PatTup _) defe) ->
+--                             let (d, nexpr) = desugarPatTup rdefs (A.Match (A.Id (Identifier ("_temp_id_" ++ show n))) [(p,ex)]) (n+1) in
+--                                 ((A.DefPat (A.PatId (fromString ("_temp_id_" ++ show n))) defe) : d, nexpr)
+--                         fndef@(A.DefFn _ _ _ _) -> first (desugarDef fndef:) $ desugarPatTup rdefs e (n+1)
+--                         _ -> first (def:) $ desugarPatTup rdefs e (n+1)
+
+desugarTupPat :: A.Program -> Compiler.Pass A.Program
+desugarTupPat p = return $ magicSubst p
+
+magicSubst :: (Data a) => a -> a
```
j-hui, line 43: Rename `magicSubst` (and others) to `desugarTup` or something related to this module name

src/Front/DesugarPatTup.hs
```diff
@@ -0,0 +1,74 @@
...
+--             where desugarPatTup [] ex _ = ([],ex)
+--                   desugarPatTup (def:rdefs) ex n =
+--                     case def of
+--                         (A.DefPat p@(A.PatTup _) defe) ->
+--                             let (d, nexpr) = desugarPatTup rdefs (A.Match (A.Id (Identifier ("_temp_id_" ++ show n))) [(p,ex)]) (n+1) in
+--                                 ((A.DefPat (A.PatId (fromString ("_temp_id_" ++ show n))) defe) : d, nexpr)
+--                         fndef@(A.DefFn _ _ _ _) -> first (desugarDef fndef:) $ desugarPatTup rdefs e (n+1)
+--                         _ -> first (def:) $ desugarPatTup rdefs e (n+1)
+
+desugarTupPat :: A.Program -> Compiler.Pass A.Program
```
j-hui, line 40: Please be consistent about `desugarTupPat` vs `desugarPatTup`

src/Front/DesugarPatTup.hs
```diff
@@ -0,0 +1,74 @@
...
+substExpr (A.Let defs e) = A.Let ndefs $ rese
+    where desugarPatTup [] ex _ = ([],ex)
+          desugarPatTup (def:rdefs) ex n =
+                    case def of
+                        (A.DefPat p@(A.PatTup _) defe) ->
+                            let (d, nexpr) = desugarPatTup rdefs (A.Match (A.Id (fromString ("_temp_id_" ++ show n))) [(p,ex)]) (n+1) in
+                                ((A.DefPat (A.PatId (fromString ("_temp_id_" ++ show n))) defe) : d, nexpr)
+                        _ -> first (def:) $ desugarPatTup rdefs e (n+1)
+          (ndefs, rese) = desugarPatTup defs e (0::Int)
+substExpr e = e
```
j-hui, line 74: Add newline to end of file

j-hui (approved):
This looks great!

Please address remaining TODOs + resolve merge conflicts before merging in.

Issue #141, opened 2023-02-20
Memory leak due to wrong main function signature


as0501:
As discussed in today's weekly meeting, we discovered that the following program causes a memory leak because the main function is missing the `cin` argument:

```
main ( cout : & Int ) -> () =
//main : &Int -> & Int -> () =
  //fun (cint : &Int) (cout : &Int)
    after 10 , (cout : & Int) <- 65
    wait (cout : & Int)
```
We commented out the lines of code that would make the program correctly print out the character 'A'. 

Here's the error produced in runtests.log:
```
sizeof(struct ssm_mm) = 4
page size 4096
pages allocated 2
objects allocated 4
objects freed 3
live objects 1
4 pools
pool   0: pages   0  block-size    16  free-blocks     0
pool   1: pages   1  block-size    64  free-blocks    63
pool   2: pages   1  block-size   256  free-blocks    16
pool   3: pages   0  block-size  1024  free-blocks     0

FAILED: 1 live objects leaked at the end
###### FAILED
```
It's not desirable for users to encounter a memory leak when using the wrong function signature for main, and they should be notified through a different type of error with a communicative error message.

PR #142, opened 2023-02-20
Add test case for memory leak on wrong main function signature

as0501:
Here is the test case for the issue as discussed in #141


j-hui:
Summary from before: the problem with this was that the sslang program was only expecting one argument, but the caller (the runtime stub) gave it two, so sslang never dropped the second argument.

j-hui:
I'm not really sure what the solution is here. The fact that arguments are passed to main as argument is already not very great, and we would probably want to change that later anyway.

PR #144, opened 2023-02-24
Implement pattern matching anomaly detection in the IR

cyoon1729:
This PR re-implements the pattern matching anomaly detection algorithm ([by Luc Maranget](http://moscova.inria.fr/~maranget/papers/warn/warn.pdf)) on the IR. The same algorithm was previously implemented to perform on the AST, but was decided to be re-implemented in the IR, together with the type checker. 

### Current status:
All test cases pass


LeoQiao18:
@cyoon1729 Can you just add a few more test cases? For example, we need a test case for "useless" pattern checking.

LeoQiao18 (approved):
Looks good to me. Nice work @cyoon1729 !

j-hui:
@cyoon1729 can you merge this in? I don't think it'll break my branch that badly.

j-hui (approved):

PR #145, opened 2023-02-24
Constraint IR Pretty Printer

jar2333:
Draft PR


jar2333:
tests are failing right now because the constraint AST is being dumped always, will make it depend on option flag when it's all done

jar2333:
have most of the skeleton needed for pretty printing the IR, I just have to look into the prettyprinter package now

leoqiao18:
@jar2333 Could you update the `Typechecking.hs` module to generate the pretty-print `Doc` only when the flag is provided?

j-hui (approved):
Fix up the regression test suite with those silly -fail test cases, and then :ship: it!

j-hui:
@jar2333 I updated the failing test error message. Feel free to squashé & mergé whenever!

PR #152, opened 2023-04-24
Support recursion in lambda lifting pass

yiming-fang:
This PR adds supports for (mutual) recursion in lambda lifting, fixing #111.


j-hui (approved):
Awesome, nice work!

j-hui:
Fix #111

PR #156, opened 2023-04-28
Remove Inliner Pass while fixing bugs

EmilySillars:
To make sure the buggy inliner does not interfere with other folks' development, let's remove it from being called in IR.hs (for now)


j-hui (approved):
Sgtm

yiming-fang (approved):

EmilySillars:
Subsumed into #152

PR #157, opened 2023-04-28
Add tuple return values for par exprs in codegen

EmilySillars:
This PR modifies code generation so that the return values of par expressions are saved in a 2-tuple.
- [tests/par-ret-vals.ssl tests](https://github.com/ssm-lang/sslang/blob/generate-ret-vals-for-par-exprs/regression-tests/tests/par-ret-vals.ssl) this functionality.
- type error thrown when a par expression takes more than two arguments (which is good because codegen does not support par expressions with more than 2 args)
- I believe this PR is ready to merge, but any feedback would be greatly appreciated!


src/Codegen/Codegen.hs
```diff
@@ -809,12 +811,31 @@ genPrim I.Par procs _ = do
...
+        onHeap <- getsDCon dconOnHeap dcon
+        unless onHeap $ do
+          fail $ "Cannot handle packed fields yet, for: " ++ show dcon
+        construct <- getsDCon dconConstruct dcon
+        destruct <- getsDCon dconDestruct dcon
+        tmp <- genTmp dty
+        let alloc = [[citem|$exp:tmp = $exp:construct;|]]
+            initField y i = [citem|$exp:(destruct i tmp) = $exp:y;|]
+            initFields = zipWith initField [ret0, ret1] [0 ..] -- puts first two return vals in a 2-tuple for now
+        return (tmp, alloc ++ initFields)
```
EmilySillars, line 832: creates a 2-tuple with the first two ret vals of par as its arguments

regression-tests/tests/foobar1.ssl
```diff
@@ -1,3 +1,6 @@
+type Pair2 a b
```
j-hui, line 1: Why was this not there before? And is it being used?

src/Codegen/Codegen.hs
```diff
@@ -38,7 +38,9 @@ import qualified Language.C.Syntax as C
 import qualified Common.Compiler as Compiler
 import Common.Identifiers (fromId, fromString)
 
-import Control.Monad (foldM, unless)
+import Control.Monad (foldM, unless, when)
+
+-- import Control.Monad (foldM, unless)
```
j-hui, line 43: Remove the commented-out import

src/Codegen/Codegen.hs
```diff
@@ -809,12 +811,31 @@ genPrim I.Par procs _ = do
         return (ret, fnStms ++ argStms, appStms)
       apply (e, _) = do
         fail $ "Cannot compile par with non-application expression: " ++ show e
+      -- given a list of par return vals and their types, wrap the return vals in a tuple
+      genParRetVal :: [C.Exp] -> [I.Type] -> GenFn (C.Exp, [C.BlockItem])
+      genParRetVal [] _ = fail "par should have 2 or more return values"
```
j-hui, line 816: nit: You can just put this as a `_` wildcard case after the `rets@(ret0:ret1:_)` case, to consolidate them

src/Codegen/Codegen.hs
```diff
@@ -809,12 +811,31 @@ genPrim I.Par procs _ = do
...
+      genParRetVal [_] _ = fail "par should have 2 or more return values"
+      genParRetVal rets@(ret0 : ret1 : _) retTys = do
+        -- TODO: given n ret vals, return an n-ary tuple
+        when (length rets /= length retTys) $ do fail "lists of return vals and types must be same length"
```
j-hui, line 820: Put the `fail` statement on the line after the `do`, i.e.,

```haskell
when (length rets /= length retTys) $ do
  fail "lists of return vals and types must be same length"
```

src/Codegen/Codegen.hs
```diff
@@ -809,12 +811,31 @@ genPrim I.Par procs _ = do
...
+        let dcon = I.tempTupleId (length rets)
+        let dty = I.tempTuple retTys
```
j-hui, line 822: Remove the redundant `let` here, should just be:

```haskell
let dcon = ...
    dty = ...
```

src/Codegen/Codegen.hs
```diff
@@ -809,12 +811,31 @@ genPrim I.Par procs _ = do
...
+        onHeap <- getsDCon dconOnHeap dcon
+        unless onHeap $ do
+          fail $ "Cannot handle packed fields yet, for: " ++ show dcon
+        construct <- getsDCon dconConstruct dcon
+        destruct <- getsDCon dconDestruct dcon
+        tmp <- genTmp dty
+        let alloc = [[citem|$exp:tmp = $exp:construct;|]]
+            initField y i = [citem|$exp:(destruct i tmp) = $exp:y;|]
+            initFields = zipWith initField [ret0, ret1] [0 ..] -- puts first two return vals in a 2-tuple for now
```
j-hui, line 831: Why only the first two values? Why not `zipWith initField rets [0..]`?

j-hui (commented):
Here's a quick review! But I have the following questions:

- Why isn't codegen for more than 2 tuples supported? It doesn't seem like the logic is any different from regular tuples.
- Why do we need to define `type Pair2 a b` in the source code, and why is it written as that (instead of `(a, b)`)?

Otherwise, the code looks really good, but those questions need to be answered/addressed before we can merge this in.

j-hui:
FYI 8d50b53b3765a8c3ccdc34f06ae24e7d6854e617 merged in support for tuples without having to define the `PairN` type. If you'd like, I can take are of integrating this and merging this in!

PR #158, opened 2023-04-29
Use symbol table to generate fresh names in InsertRefCounting

j-hui:
- Add name mangling pass to make VarIds globally unique
- Use slightly more esoteric mangling to avoid clashes
- Actually update the varNames map
- Make LambdaLift use manglenames
- Fix Lambda lift pass
- Fix things after different binder structure
- Still broken, waiting for types to appear in binders
- Ok ok ok ok
- Get it compiling
- Passtests
- Some small fixes
- Use symbol table to generate fresh names in InsertRefCounting


j-hui:
actually, no point in breaking up this into a separate commit. This commit will be rolled in together with #120

PR #160, opened 2023-05-04
Constraint Pretty-Print: WriterT transformer for TC monad

jar2333:
Uses a WriterT monad transformer on the TC monad stack to implement pretty-printing of constraint information regardless of success or failure of type-checking. Fully working, some points to note:
- I did not modify the Monad.hs file to export the WriterT monad's `tell` and `when` functions. This could be done to avoid having to import them separately in code using the TC monad, perhaps through wrapper functions.
- I combine the pretty-printed output and any errors that result through a simple string concatenation. If the errors are pretty-printed later, this code would have to change. Let me know if this has to be modified.


src/IR/Constraint/Typechecking.hs
```diff
@@ -29,50 +29,64 @@ import Prettyprinter (
...
+  I.Program Can.Annotations -> Bool -> Compiler.Pass (I.Program Can.Type)
+typecheckProgram pAnn False = do
+  let (result, _) = unsafeTypecheckProgram pAnn False
+  case result of
+    Left e -> Compiler.throwError e
+    Right pType -> return pType
+typecheckProgram pAnn True = do
+  let (result, pp) = unsafeTypecheckProgram pAnn True
+  case result of
+    Left e -> Compiler.dump $ show pp ++ "\n\n" ++ show e
```
leoqiao18, line 48: Ideally, these should be 2 separate `Compiler.Error` cases (`pp` is a `Dump`, and `e` is a `TypeError`). However, the `Compiler.Pass` monad only supports throwing 1 error through the `ExceptT`. @j-hui , is what @jar2333 doing here okay (i.e. putting the type error together with a dump)?

j-hui: It's not the cleanest way to do this, but I'm not opposed to it. What does the output look like, with and without a type error? e.g., what is the output for the following programs:

```
f x = x
```

```
f x = f
```

jar2333: `f x = x`
```
let rec ['ir_t0]
  f : 'ir_t0
given
  exists ['ir_t1, 't4]
    def
      x : 'ir_t1
    in
      local x <: 't4
    ('ir_t1 -> 't4) = 'ir_t0
in
  true

--------------------

f (x: 'ir_t1): 'ir_t0 = x

--------------------

'ir_t0 = ('t4 -> 't4)
'ir_t1 = 't4
'ir_t2 = 't4
'ir_t3 = ('t4 -> 't4)
```

vs

`f x = f`
```
let rec ['ir_t0]
  f : 'ir_t0
given
  exists ['ir_t1, 't4]
    def
      x : 'ir_t1
    in
      local f <: 't4
    ('ir_t1 -> 't4) = 'ir_t0
in
  true

--------------------

f (x: 'ir_t1): 'ir_t0 = f

TypeError (ErrorMsg "Infinite type for variable f, which has the infinite type: Type -> [FlexVar 'ir_t1,Infinite]")
```

I thought it worked fine since currently errors are just printed using the show instance regardless, although of course it can be cleaner and/or open to revision

j-hui: Ok, I think that looks good to me! Maybe add a comment making a note of the fact that you're sort of "misusing" the ExceptT slightly.

j-hui (approved):
GitHub tells me there are some redundant imports warnings that you should probably clean up, but aside from that, LGTM!

PR #162, opened 2023-05-17
Support tuple syntax without having to declare PairN

j-hui:


j-hui:
As long as tests pass, I'm going to merge this in without review because I desperately need this feature for a paper submission 🥲

